PR #15, opened 2025-12-02
Enable Local File Sources in Provider Discovery

robotdad:
When using the installed version of Amplifier (via pip install or uv tool install), users
  could only configure providers in ~/.amplifier/settings.yaml with git sources. Local file
  paths were not supported in the provider discovery and configuration system, making it
  difficult to:

  - Test local provider modifications without publishing to git
  - Use custom local providers alongside the installed Amplifier CLI
  - Develop providers locally while using the production CLI

  This limitation only affected the global settings configuration - local project configs worked
   fine with file sources.

  Solution

  This PR adds support for local file sources in the provider discovery and configuration
  system:

  1. Provider Discovery: The module source resolver now properly handles local file paths (e.g.,
   file:///path/to/provider, /absolute/path, ./relative/path)
  2. Configuration Validation: Local file sources are now recognized as valid alongside git
  sources in settings.yaml
  3. None Check Safety: Added defensive check for amplifier_core.__file__ to prevent
  AttributeError in certain import contexts

### DIFF
--- a/amplifier_app_cli/commands/module.py
+++ b/amplifier_app_cli/commands/module.py
@@ -587,6 +587,9 @@ def _run_behavioral_tests(module_path: str, module_type: str) -> bool:
     try:
         import amplifier_core
 
+        if amplifier_core.__file__ is None:
+            console.print("[red]amplifier-core __file__ not available - cannot locate behavioral tests[/red]")
+            return False
         core_path = Path(amplifier_core.__file__).parent
         test_file = core_path / "validation" / "behavioral" / f"test_{module_type}.py"
 


bkrabach:
Thank you for the contribution! This PR addressed an important need.

After investigation, we found that this functionality was implemented in main on December 5, 2025 (commit `5b89922`). The feature has been further refined with several follow-up commits for edge cases like auto-installing dependencies for local providers.

Local file sources now work correctly:
- `amplifier provider list` shows providers from local file sources
- Settings.yaml `sources:` section supports local paths (`./path`, `/absolute/path`, `file://path`)
- `amplifier source add` handles local paths properly

Closing this PR as the functionality is already in main. Thank you for identifying this need!

Issue #21, opened 2025-12-03
[High] Interactive Mode Enhancements


marklicata:
## Summary
Enhance interactive mode with input prefixes (`!` for bash, `#` for memory), vim keybindings, multi-line input support, and improved keyboard shortcuts.

## Current State
Amplifier CLI interactive mode has basic prompt-toolkit integration but lacks:
- Input prefixes for quick actions
- Vim mode for efficient editing
- Multi-line input with backslash continuation
- Advanced keyboard shortcuts

### 1. Input Prefixes

| Prefix | Action | Example |
|--------|--------|---------|
| `!` | Execute bash directly | `!git status` |
| `#` | Append to memory | `# Remember: use pytest-asyncio` |
| `/` | Slash command (exists) | `/help` |
| `?` | Quick help/search | `?how do I run tests` |

```
amplifier> !ls -la
drwxr-xr-x  10 user  staff   320 Jan 15 10:30 src
-rw-r--r--   1 user  staff  1024 Jan 15 10:25 README.md

amplifier> # The project uses FastAPI 0.100+
[Added to session memory]

amplifier> ?pytest fixtures
[Quick search: "pytest fixtures" - showing brief answer]
```

### 2. Vim Mode

Enable vim keybindings in interactive mode:

```yaml
# In amplifier.yaml
interactive:
  editing_mode: "vi"  # Options: "emacs" (default), "vi"
```

Or via CLI:
```bash
amplifier --vim
amplifier -V  # Short flag
```

Vim mode features:
- Normal/insert mode switching
- Motion commands (w, b, e, 0, $, etc.)
- Delete/change/yank operations
- Search within line
- Visual mode for selection

### 3. Multi-Line Input

Use backslash `\` at end of line to continue:

```
amplifier> Create a Python function that \
... takes a list of numbers and \
... returns the sum of even numbers

[Processing multi-line input...]
```

Or use triple backticks for code blocks:

```
amplifier> ```
... def calculate_sum(numbers):
...     return sum(n for n in numbers if n % 2 == 0)
... ```
Please review this function.

[Processing with code block...]
```

### 4. Keyboard Shortcuts

| Shortcut | Current | Proposed |
|----------|---------|----------|
| `Ctrl+C` | Cancel | Cancel current (exists) |
| `Ctrl+D` | Exit | Exit if empty (exists) |
| `Ctrl+L` | - | Clear screen |
| `Ctrl+R` | - | Search history |
[...]

PR #186, opened 2026-05-19
fix(CommandProcessor): emit mode lifecycle events from _handle_mode

colombod:
## Problem

`CommandProcessor._handle_mode` activates and deactivates modes by directly mutating `session_state["active_mode"]` without emitting the corresponding lifecycle events (`mode:activated`, `mode:changed`, `mode:cleared`).

This means `hooks-mode`'s `handle_mode_activated` handler is never called, so `RuntimeOverlay.apply()` never runs, and `contributes.agents` / `contributes.skills` / `contributes.context` declared in a mode's `contributes:` block are silently skipped.

The `mode` tool and orchestrator path emit these events correctly via the hook system. `_handle_mode` (the interactive CommandProcessor shortcut path — used when a human types `/mode-name` in an interactive session) was the only path that did not.

## Symptom

In an interactive session, typing `/context-intelligence` activates the mode body and context injection correctly (that path reads `session_state["active_mode"]` directly). But agents declared under `contributes.agents:` are never mounted, so delegating to them fails:

```
ValueError: Agent 'context-intelligence:context-intelligence-design-facilitator'
            not found in configuration
```

This is the root cause of amplifier-support#233.

## Fix

After each `session_state["active_mode"]` mutation in `_handle_mode`, emit the appropriate hook event so the full lifecycle chain runs:

| Path | State change | Event emitted |
|------|-------------|---------------|
| `/mode off` | active → None | `mode:cleared` |
| explicit `on` (no prior mode) | None → new | `mode:activated` |
| explicit `on` (switching modes) | old → new | `mode:changed` |
| explicit `off` | active → None | `mode:cleared` |
| toggle (no prior mode) | None → new | `mode:activated` |
| toggle (switching modes) | old → new | `mode:changed` |
| toggle (already active) | active → None | `mode:cleared` |

## Test coverage

- **897 tests pass** (1 pre-existing failure in `test_session_spawner_subprocess.py::test_subprocess_param_routes_to_subprocess` unrelated to this change — failing on `main` before this PR).
- All mode-specific tests pass: `tests/test_modes_display.py`, `tests/test_mode_trailing_prompt.py`.
- **DTU end-to-end validation** (DTU `ci-appcli-fix`): confirmed the full pipeline — `mode:activated` → `handle_mode_activated` → `overlay.apply()` → `contributes.agents` mounted → `delegate:agent_completed` for `context-intelligence:context-intelligence-design-facilitator`.

```jsonl
{"event": "mode:activated", "data": {"mode": "context-intelligence", "name": "context-intelligence", ...}}
{"event": "delegate:agent_completed", "data": {"agent": "context-intelligence:context-intelligence-design-facilitator", "success": true, ...}}
```

## Companion fix

`amplifier-bundle-context-intelligence` (already on `main`) has the companion fix: `contributes.agents` keys must be fully namespaced (`context-intelligence:context-intelligence-design-facilitator`) to match the agent name used by the delegate tool's lookup. Without that fix, even with this PR, the agent key in `coordinator.config` would not match the lookup key.

### DIFF
--- a/amplifier_app_cli/main.py
+++ b/amplifier_app_cli/main.py
@@ -626,15 +626,15 @@ class CommandProcessor:
         # /mode off - clear any active mode
         if args == "off":
             if current_mode:
+                # Emit mode:cleared BEFORE state mutation so hooks see the old state
+                await self.session.coordinator.hooks.emit(
+                    "mode:cleared", {"name": current_mode, "previous_mode": current_mode}
+                )
                 session_state["active_mode"] = None
                 # Reset warnings in mode hooks if present
                 mode_hooks = session_state.get("mode_hooks")
                 if mode_hooks and hasattr(mode_hooks, "reset_warnings"):
                     mode_hooks.reset_warnings()
-                # Emit mode:cleared so hooks-mode revokes contributes.* (overlay.revoke)
-                await self.session.coordinator.hooks.emit(
-                    "mode:cleared", {"name": current_mode, "previous_mode": current_mode}
-                )
                 return f"Mode off: {current_mode}"
             return "No mode active"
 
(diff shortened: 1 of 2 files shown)


bkrabach (commented):
This is the right shape and the right layer. PR #21 was solving the symptom in the wrong place; this solves the cause in the right place. All seven non-trivial state transitions covered, including the deactivate paths the original PR didn't touch. The end-to-end DTU chain (`mode:activated` → `handle_mode_activated` → `overlay.apply()` → `delegate:agent_completed`) is the right kind of validation for a behavioral fix.

Two contract nits and one test ask before merge. None blocking — the fix is correct as-is. These are about hardening the contract so two emitters stay aligned.

### 1. Payload shape diverges from tool-mode

`modules/tool-mode/__init__.py:358-374` emits `mode:activated` with a richer payload:

```python
{"name": name, "mode": name,
 "description": mode_def.description,
 "default_action": mode_def.default_action}
```

This PR emits:

```python
{"name": mode_name, "mode": mode_name}
```

The hooks-mode handlers don't care — they re-derive from `discovery.find()`. But third-party observers subscribed to `mode:activated` will see different fields depending on whether the user typed `/foo` (this PR's path) or the LLM called the `mode` tool. Suggest matching tool-mode's payload — `description` is already on hand via `_lookup_mode_description`; `default_action` is one more discovery lookup. Same alignment for `mode:changed` and `mode:cleared`.

### 2. Emit order is inverted vs tool-mode

`tool-mode/__init__.py:384-389` emits BEFORE mutating `session_state["active_mode"]`. The comment there is explicit:

> "emit() is awaited BEFORE the state writes... if emit() raises, observable state stays consistent with the return value."

This PR does the opposite — mutates first, then emits. Practically fine today because hooks-mode handlers read from the event payload. But the handlers also have a `session_state["active_mode"]` fallback for missing canonical keys; if anything ever hits that fallback, it will see the post-mutation value instead of the pre-mutation one. Suggest aligning by swapping the order:

```python
await self.session.coordinator.hooks.emit("mode:activated", {...})
session_state["active_mode"] = mode_name
```

Same code, different order. Brings `CommandProcessor` into contract symmetry with `tool-mode`.

### 3. Add unit tests for the seven emits

DTU is good evidence for "the fix works today." It does not guard against "a future refactor silently removes one of these emits." Cheap test, modeled on existing CommandProcessor patterns:

```python
async def test_handle_mode_emits_activated_on_explicit_on(): ...
async def test_handle_mode_emits_changed_on_explicit_on_switching(): ...
async def test_handle_mode_emits_cleared_on_explicit_off(): ...
async def test_handle_mode_emits_cleared_on_bare_off(): ...
async def test_handle_mode_emits_activated_on_toggle_no_prior(): ...
async def test_handle_mode_emits_changed_on_toggle_switching(): ...
async def test_handle_mode_emits_cleared_on_toggle_already_active(): ...
```

Seven tests, one per branch. Each one prevents a specific regression in this file.

### 4. Close amplifier-bundle-modes#21 as superseded

That PR's late-bind block in `handle_provider_request` becomes unnecessary with this fix, and incomplete (it didn't cover the deactivate path). Worth referencing as superseded and closing explicitly.

colombod:
Thanks for the thorough review, @bkrabach. All four items addressed — pushed as `8bde063` on top of the original `4a2611a`.

### Issue 1 — Payload shape ✅

`mode:activated` and `mode:changed` now carry the same fields as tool-mode. `mode:cleared` was already minimal and correct (no change needed).

```python
# mode:activated (no prior mode)
{
    "name": mode_name, "mode": mode_name,          # canonical + legacy
    "description":    mode_def.description,
    "default_action": mode_def.default_action,
    "safe_tools":     mode_def.safe_tools,
    "warn_tools":     mode_def.warn_tools,
    "confirm_tools":  mode_def.confirm_tools,
    "block_tools":    mode_def.block_tools,
}

# mode:changed (switching modes)
{
    "old": _prev, "new": mode_name,                # canonical
    "from_mode": _prev, "to_mode": mode_name,      # legacy compat
    "description":    mode_def.description,
    "default_action": mode_def.default_action,
    "safe_tools": ..., "warn_tools": ..., ...      # same as above
}
```

When `discovery` is `None` (no mode definition available), a minimal fallback payload is emitted — same keys as before for those paths.

### Issue 2 — Emit order ✅

All five emit calls moved to fire **before** the corresponding `session_state["active_mode"]` mutation. Matches the explicit tool-mode contract: *"emit() is awaited BEFORE the state writes… if emit() raises, observable state stays consistent with the return value."*

### Issue 3 — Unit tests ✅

Added `tests/test_handle_mode_events.py` with seven tests, one per state-transition path:

| Test | Event | Path |
|---|---|---|
| `test_handle_mode_emits_cleared_on_bare_off` | `mode:cleared` | `/mode off` |
| `test_handle_mode_emits_activated_on_explicit_on` | `mode:activated` | explicit on, no prior |
| `test_handle_mode_emits_changed_on_explicit_on_switching` | `mode:changed` | explicit on, prior exists |
| `test_handle_mode_emits_cleared_on_explicit_off` | `mode:cleared` | explicit off |
| `test_handle_mode_emits_activated_on_toggle_no_prior` | `mode:activated` | toggle, no prior |
| `test_handle_mode_emits_changed_on_toggle_switching` | `mode:changed` | toggle, prior exists |
| `test_handle_mode_emits_cleared_on_toggle_already_active` | `mode:cleared` | toggle off |

Each test asserts three things:
1. The correct event name fired
2. The full payload shape (all canonical + legacy-compat keys present with correct values)
3. **Emit fired before mutation** — captures `session_state["active_mode"]` inside the event handler at call time and asserts it still holds the pre-mutation value

### Issue 4 — Modes bundle PR #21 ✅

Already closed before this review with an explanation pointing here as the correct fix.

---

**Full test run: 952/952 pass** (1 pre-existing failure in `test_session_spawner_subprocess` on `main` before this PR, unrelated).

bkrabach (approved):
All four items resolved in `8bde063`. Approving.

- **Payload shape:** full 8-key match with tool-mode (I underspecified in my pointer — the real target included the tool-policy fields. Match confirmed against `modules/tool-mode/__init__.py:357-381`).
- **Emit order:** every site flipped to fire-before-mutate. tool-mode's contract comment preserved.
- **Tests:** the third assertion in each test captures `session_state["active_mode"]` inside the event handler and verifies it still holds the pre-mutation value. That encodes the emit-order contract, not just emit-presence — stronger guard than asked for.
- **#21:** closed before this review even posted. Saved us the step.

952/952 passing. Ship it.

PR #217, opened 2026-07-06
fix: restore cumulative session cost on resume

kenotron-ms:
## Summary

Restores cumulative session LLM cost when a session is resumed. The cost counter currently resets to zero on resume, even though the per-turn cost display remains correct.

**Issue:** microsoft-amplifier/amplifier-support#284

## Root Cause

Each provider module accumulates session cost in-memory inside its `mount()` closure and registers a contributor function that adds the accumulated value to the kernel `session.cost` channel. On resume, the provider re-mounts with a zeroed accumulator, so the running session total restarts from zero.

Per-turn cost display remains correct because it reports only the deltas from turns executed in the current session — but the cumulative counter, which should preserve and build on prior turns, loses history.

## The Fix

Introduced a cost-history bridge (~30 lines, zero changes to provider modules or core):

**New file: `amplifier_app_cli/cost_history.py`**
- `sum_prior_cost_usd(events_path)` streams the session's persisted `events.jsonl` line-by-line (memory-safe) and sums `data.usage.cost_usd` from all `llm:response` events.
- `restore_session_cost(coordinator, session_id, events_path)` re-seeds the resumed session's coordinator by registering a synthetic `session.cost` contributor named `history:<session_id>` that contributes the prior session cost. Mirrors the existing `amplifier_foundation.bridge_child_cost` register_contributor pattern.
- Best-effort, never raises; silently no-ops on missing/malformed events or zero prior cost.

**Edit: `amplifier_app_cli/session_runner.py`**
- Added Step 7.5 in `create_initialized_session()` right after transcript restore.
- Guarded by `config.is_resume` so it only runs on resume.
- Computes events path as `SessionStore().base_dir / session_id / "events.jsonl"`.

**No double-counting:** Register_contributor appends, and the fresh provider accumulator only counts turns executed after resume, so history + live contributions together reproduce the true cumulative total.

## Testing

- `uv run pytest tests/test_cost_history.py` → **11 tests passed**
  - Summing cost from events
  - Summing with malformed/null lines (robust)
  - Ignoring non-llm:response events
  - Contributor registration  
  - Missing/zero cost files handled gracefully
  - Error-swallowing (never raises)
  - Resume scenario: no double-count between history + live

- `python_check` (ruff format+lint, pyright) → **clean** on both new files

- Cost field path confirmed empirically: `data.usage.cost_usd` (stringified decimal) in real `llm:response` events from session.

**Pre-existing note:** `tests/test_cost_bridge.py` fails to collect in a clean `uv sync` env — the installed `amplifier_foundation` 1.0.0 does not export `bridge_child_cost` / `sum_cost_usd`, indicating a pre-existing lockfile/env drift unrelated to this change. The new test is self-contained and does not depend on those exports.

## Follow-up / Out of Scope

The secondary in-memory accumulator `state["prev_total"]` in the `hooks-streaming-ui` module (a separate repository) also resets on resume. On the first post-resume turn, the per-turn delta display may transiently include restored cost (appearing to show an unexpectedly large first turn). Fixing the per-turn display presentation requires a change in that hook module and is out of scope for this app-cli fix, which restores the cumulative session total (the issue's primary complaint).

---

### DIFF
--- a/amplifier_app_cli/cost_history.py
+++ b/amplifier_app_cli/cost_history.py
@@ -1,0 +1,134 @@
+"""Restore cumulative session cost on resume (issue #284).
+
+Session LLM cost is accumulated **in-memory** in each provider module's
+``mount()`` closure (a ``_totals`` dict) and contributed to the kernel's
+``session.cost`` channel. When a session is resumed the provider re-mounts with
+that accumulator back at zero, so the cumulative session-cost counter restarts
+from zero. (Per-turn cost still displays correctly; only the running session
+total is lost.)
+
+This module reads the prior cumulative cost from the session's persisted
+``events.jsonl`` -- every ``llm:response`` event carries ``data.usage.cost_usd``
+-- and re-seeds the running total by registering a synthetic ``session.cost``
+contributor on the *resumed session's own coordinator*. This mirrors the
+``register_contributor`` pattern already used by
+``amplifier_foundation.bridge_child_cost`` to bridge child-session cost into a
+parent.
+
+Design notes:
+- ``register_contributor`` APPENDS (the kernel never overwrites on duplicate
+  name), and ``collect_contributions`` sums every registered contributor. The
+  fresh per-mount provider accumulator starts at zero on resume and only counts
+  turns executed *after* the resume, so the historical contributor and the live
+  provider contributor never double-count the same spend.
+- Everything here is best-effort and must never break session startup: missing
+  or corrupt event files simply yield no restored cost.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from decimal import Decimal
+from decimal import InvalidOperation
+from pathlib import Path
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+# Channel + event/field names defined by the kernel and provider modules.
+SESSION_COST_CHANNEL = "session.cost"
+_LLM_RESPONSE_EVENT = "llm:response"
+
+
+def sum_prior_cost_usd(events_path: Path) -> Decimal | None:
+    """Sum ``cost_usd`` across every ``llm:response`` event in ``events_path``.
+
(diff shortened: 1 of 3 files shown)


bkrabach:
## DTU validation — verified end-to-end

Ran this against a live DTU (Incus container, real Anthropic provider) rather than relying on the unit tests alone:

1. Started a fresh session, ran 3 real turns, confirmed cost accumulating in `events.jsonl` (`$0.00665745 → $0.01340490 → $0.02024235`).
2. Resumed the same session twice via real `amplifier run --resume <id>` invocations — confirmed `config.is_resume` correctly triggers step 7.5.
3. Ran `sum_prior_cost_usd()` / `restore_session_cost()` directly against the real, provider-generated `events.jsonl` — sums matched the manual total exactly at each step, no double-counting against the live per-mount provider accumulator.
4. Verified graceful handling of: missing `events.jsonl`, empty file, no `llm:response` events, and a corrupt JSON line mixed with a valid one — all `None`/no-op, no exceptions, matching the "never raises" contract.

One thing worth calling out for reviewers (not a blocker — you already scoped this correctly in the "Follow-up / Out of Scope" section): confirmed independently that the visible "Turn: $X | Session: $Y" readout comes from a separate `hooks-streaming-ui` module with its own pre-existing display bug. This PR correctly fixes the underlying `session.cost` data/channel; it does not by itself change what's printed on screen.

Also caught and corrected a false alarm on my end: an initial `git diff main pr-217` showed what looked like an unrelated 427-line test deletion — turned out to be my local `main` having drifted past the PR's actual merge-base, not a real discrepancy. Confirmed via `git diff <merge-base> pr-217` that the true diff is exactly the 3 files / 319 additions / 0 deletions this PR claims.

Nice fix — clean, well-scoped, and it held up under real usage, not just review.

PR #210, opened 2026-07-02
feat: steering input UX + DRY input-reuse

bkrabach:
Mid-turn steering — type a message while the agent is working and it's acted on between tool rounds (redirect it, add context, @-mentions) instead of waiting for the turn to finish. Ships across 3 repos; proven end-to-end in a Digital Twin Universe.

## What this PR does

Concurrent bottom-pinned steering prompt during a turn (self-healing queued-count badge, hidden label, honest acks); Ctrl-C cancels cleanly with no REPL crash; steered input flows through the SAME processing as normal REPL input — @-mention expansion works, and slash commands are recognized and rejected mid-turn (reject-all for now; a `/mode` mid-turn whitelist and a FollowUpQueue are intentionally deferred); plus `is_composing` wiring for the streaming throttle in the companion hooks-streaming-ui PR. Full steering test suite + python_check clean. 

DTU-proven: mid-turn redirect, `@file` expansion, `/help` rejected (not injected raw), Ctrl-C graceful.

## Companion PRs

- [amplifier-module-loop-streaming: mid-turn steering mechanism](#) (https://github.com/microsoft/amplifier-module-loop-streaming/pull/29)
- [amplifier-module-hooks-streaming-ui: streaming that coexists with a pinned prompt](#) (link after all three exist)

PR #219, opened 2026-07-07
fix: resolve terminal echo corruption and mid-turn Ctrl-C cancellation regressions

bkrabach:
## Summary

This PR fixes two real, live regressions introduced by the steering feature (PR #210):

1. **Terminal echo corruption after exit** — Keystrokes sometimes weren't echoed after exiting an interactive session due to orphaned asyncio tasks holding the terminal in raw mode
2. **Ctrl-C during a turn does nothing** — Physical Ctrl-C keypresses during active turns were silently swallowed with no feedback or cancellation

Both regressions have been fixed, tested with integration test suites, and verified live in a DTU.

---

### Root Cause
`SteeringInputManager.run()` could orphan a child asyncio task holding the terminal in raw mode. Across multi-turn sessions, these orphans accumulate. On Python's non-deterministic shutdown-sweep order, the terminal could be left corrupted (ECHO/ICANON disabled) after a normal exit (zero Ctrl-C, zero signals).

### The Fix
`run()`'s `finally:` block now unconditionally cancels and awaits any live child task before completing, ensuring every exit path closes with the same guarantee: terminal state restored.

**Commits:** 57daf75, 2cffc88  
**Tests:** `tests/test_terminal_echo_integration.py` — 7 scenarios including the headline `multi_orphan` case (4 sequential turns, zero Ctrl-C, zero signals, terminal corrupted before fix / healthy after)  
**Evidence:** Full suite green (1017 passed, same 14 pre-existing unrelated failures unchanged); integration test proves terminal state recovery; verified live in DTU against installed package.

---

### Root Cause
prompt_toolkit's `raw_mode()` (held for most of an active turn by the steering prompt) disables `ISIG`, so a Ctrl-C keypress never generates an OS-level SIGINT while that prompt has focus. The raw byte arrives only inside `steering_input.py` and was consumed entirely with no forwarding to the session's cancellation token.

### The Fix
The `_CtrlCInterrupt` and `KeyboardInterrupt` handlers in `run()`'s poll loop now forward to `session.coordinator.cancellation` with the **same graceful-then-immediate escalation** and **same visible console messages** as the existing OS-signal path (`sigint_handler` in main.py). Ctrl-C now behaves identically regardless of which path catches it.

**Commits:** 51a55cf, 1bf25be  
**Tests:** `tests/test_ctrlc_functional_integration.py` — 4 scenarios proving Ctrl-C now actually escalates graceful→immediate and stops the task  
**Evidence:** Full suite green; functional integration test proves cancellation propagation; verified live in DTU against installed package.

---

## Testing Summary

- **1017 tests passed** across the full suite
- **14 pre-existing unrelated failures** unchanged (baseline maintained)
- **0 new failures** introduced by these fixes
- **python_check clean** — no format, lint, or type issues
- **All commits follow conventional commit style** with Amplifier co-author attribution
- **Both fixes verified end-to-end in a fresh DTU** against the actually-installed package (not just dev-machine state)
- **User personally verified both fixes live in a DTU** — confirmed they work great

---

## Known Follow-Up (Not In This PR)

A related-but-distinct race in the arbiter-abort branch of `run()`'s poll loop: an external cancellation landing at that specific `await` gets silently absorbed rather than propagated. Currently harmless only because the caller always sets a stop flag before cancelling, but not defended inside `run()` itself.

This is a **separate tracked issue** (to be filed after PR merge) rather than folded into this PR. The title will be something like: "Fragility: arbiter-abort branch in SteeringInputManager.run() can silently swallow an external cancellation"

---

### DIFF
--- a/amplifier_app_cli/approval_provider.py
+++ b/amplifier_app_cli/approval_provider.py
@@ -1,5 +1,7 @@
 """CLI approval provider for interactive user approval."""
 
+from __future__ import annotations
+
 import asyncio
 import logging
 
@@ -8,6 +10,8 @@ from amplifier_core import ApprovalResponse
 from rich.console import Console
 from rich.panel import Panel
 from rich.prompt import Confirm
+
+from .stdin_arbiter import StdinArbiter
 
 logger = logging.getLogger(__name__)
 
@@ -19,14 +23,16 @@ class CLIApprovalProvider:
     Implements ApprovalProvider protocol for CLI environments.
     """
 
-    def __init__(self, console: Console):
+    def __init__(self, console: Console, arbiter: StdinArbiter | None = None):
         """
         Initialize CLI approval provider.
 
         Args:
             console: Rich console for output
+            arbiter: Optional stdin arbiter for coordinating with steering reader
         """
         self.console = console
+        self._arbiter = arbiter
 
     async def request_approval(self, request: ApprovalRequest) -> ApprovalResponse:
         """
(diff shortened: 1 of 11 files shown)


bkrabach:
Superseded by #220 with clean cherry-picked commits

PR #239, opened 2026-07-28
fix(providers): repair all configured provider modules, not just the active one

bkrabach:
## Problem

After `amplifier update` or `amplifier reset`, only one provider module ends up installed. Every other configured provider silently fails to mount:

```
Installing provider module...
Installing provider-anthropic... ✓
...
Partial provider failure: 4/10 loaded. Missing: {'openai': 1, 'gemini': 1,
'github-copilot': 1, 'qwen-3.6': 1, 'azure-openai': 1, 'ornith': 1}.
```

The session continues in a degraded state with only a log warning. Install state correlated 1:1 with mount failures — the one installed module's 4 instances mounted, the 5 uninstalled modules' 6 instances did not.

## Root cause

Two defects in `check_first_run()`, which is the **only** provider-repair path in the CLI. `update.py` and `reset.py` contain zero references to providers — both recreate the tool venv (`uv tool install --upgrade --reinstall`, and `uninstall` + `install` respectively), wiping all editable-installed provider modules, then rely entirely on the next boot's `check_first_run()`.

**1. Regression from #238 (ad69c69).** The repair call was narrowed:

```python
-        installed = install_known_providers(config, console, verbose=True)
+        installed = ensure_provider_installed(current_provider.module_id, ...)
```

Repairing one provider cannot fix a wipe that removed all of them.

**2. Pre-existing latent bug.** The repair was gated on the active provider only — present in `ad69c69^` as well:

```python
module_installed = _is_provider_module_installed(current_provider.module_id)
if not module_installed:
    ...
```

If the active provider happened to be installed but another configured provider was missing, no boot ever repaired it. Reachable via a transient install failure (`install_known_providers` catches per-provider exceptions and continues), a hand-edited `settings.yaml`, or `amplifier provider install <one-name>`.

## Fix

Gate the repair on **every** configured provider module, and repair with `install_known_providers()`:

```python
configured_modules = _configured_provider_modules(config, current_provider.module_id)
missing = [m for m in configured_modules if not _is_provider_module_installed(m)]
if missing:
    installed = install_known_providers(config, console, verbose=True)
```

Return value stays driven by the **active** provider only — a non-active provider failing to install must never push the user into the add-a-provider prompt. Modules that cannot be repaired (neither a known provider nor carrying a `source:`) now emit a named warning rather than failing silently.

|  | Gate | Repair | Broken cases |
|---|---|---|---|
| pre-#238 | active only | all known | active present + others missing |
| #238 (current main) | active only | active only | ↑ plus full venv wipe |
| **this PR** | **all configured** | **all known** | none of the above |

## This preserves #238 — it is not a revert

#238 fixed four things. This PR touches one of them.

| # | What #238 fixed | Where | Touched here? |
|---|---|---|---|
| 1 | Pinned `source:` takes precedence over built-in defaults | `get_effective_provider_sources()` | No |
| 2 | Already-installed providers skipped, not rebuilt every run | `install_known_providers()` guard | No |
| 3 | Entry point must actually resolve (repairs dangling editable installs) | `is_provider_module_installed()` | No |
| 4 | `check_first_run()` repairs only the one missing provider | `check_first_run()` | **Replaced** |

Items 1–3 are what close #342. Item 4 was defended in #238's commit message as protecting pinned installs — but item 2, added in the same PR, already does that. `install_known_providers(force=False)` skips anything already installed, so it never overwrites a pinned build. The narrowing was redundant, and it is what caused the regression.

**Evidence:** #238's own test file `tests/test_provider_source_precedence.py` shipped 5 classes / 14 tests. Exactly one class changed here — the one pinning item 4. The other 13 pass unchanged:

```
TestEffectiveProviderSources ............ 7 PASSED   (item 1)
TestInstallKnownProvidersIdempotency .... 2 PASSED   (item 2)
TestDanglingEditableInstallIsNotSkipped . 4 PASSED   (item 3)
TestCheckFirstRunRepairsAllKnownProviders 1 PASSED   (item 4, reworked)
14 passed
```

Reverting #238 instead would reintroduce #342 (pinned sources overwritten with `@main`, every provider rebuilt each run), lose dangling-editable-install detection, and still carry the pre-existing active-only gate bug.

## Test plan

- [x] 5 new tests in `tests/test_check_first_run_installs_all_providers.py`, written first; 3 failed against pre-fix code, including one reproducing the active-installed/others-missing hole
- [x] Targeted suite: **22 passed**
- [x] Full suite: **15 failed, 1174 passed** vs `main` baseline **15 failed, 1166 passed** — identical failure list (all pre-existing, unrelated files); +8 = the new tests
- [x] `ruff check` on all touched files: clean
- [x] E2E, isolated venv + temp `HOME`, 4 provider modules configured, none installed → all 8 known providers installed, all 4 configured modules resolve, `check_first_run() -> False`
- [x] E2E, active installed + `provider-openai` removed → repair triggered, 7 reported `already installed, skipping`, only `provider-openai` installed. Confirms #238's skip-guard still protects pinned builds under real conditions.
- [x] Boot cost measured: ~6–7 ms for 4 configured modules on the happy path (`find_spec` only; no subprocess, no network), still short-circuits to `return False`

## Rollout

Existing users get this with one `amplifier update` — the update replaces app-cli, and the next launch runs the new gate and restores everything. Users currently sitting in the degraded state are repaired the same way. Fresh installs were never affected. First launch after updating will spend ~30–60s installing providers from git; this is one-time and only when something is genuinely missing.

### DIFF
--- a/amplifier_app_cli/commands/init.py
+++ b/amplifier_app_cli/commands/init.py
@@ -19,7 +19,6 @@ from ..ui.scope import (
 from ..provider_config_utils import configure_provider
 from ..provider_manager import ProviderManager
 from ..provider_env_detect import detect_provider_from_env
-from ..provider_sources import ensure_provider_installed
 from ..provider_sources import install_known_providers
 from .routing import _discover_matrix_files
 from .routing import _get_configured_provider_types
@@ -53,6 +52,48 @@ def _is_provider_module_installed(provider_id: str) -> bool:
     return is_provider_module_installed(provider_id)
 
 
+def _configured_provider_modules(
+    config: AppSettings, active_module_id: str
+) -> list[str]:
+    """Ordered unique provider module IDs configured in settings.
+
+    The active provider is always first: it is configured by definition, and
+    check_first_run()'s return value is driven by it.
+
+    Reads `config.providers` via the same accessor the rest of the install path
+    uses (`provider_sources.get_effective_provider_sources()` reads it too), so
+    install time and this check agree on what "configured" means. Malformed
+    entries are skipped rather than raised on - a hand-edited settings.yaml
+    must never prevent Amplifier from booting.
+
+    Args:
+        config: Config manager (AppSettings) to read configured providers from
+        active_module_id: Module ID of the currently active provider
+
+    Returns:
+        Ordered list of unique provider module IDs, active provider first
+    """
+    modules = [active_module_id]
+
+    try:
+        configured = config.get_provider_overrides()
+    except Exception as e:  # pragma: no cover - defensive, settings are best-effort
+        logger.debug(f"Could not read configured providers: {e}")
+        return modules
+
+    if not isinstance(configured, list):
+        return modules
+
+    for entry in configured:
+        if not isinstance(entry, dict):
+            continue
+        module_id = entry.get("module")
+        if module_id and module_id not in modules:
+            modules.append(module_id)
+
+    return modules
+
+
 def check_first_run() -> bool:
     """Check if this appears to be first run (no provider configured).
 
@@ -70,10 +111,22 @@ def check_first_run() -> bool:
     2. No surprise defaults that may not match bundle requirements
     3. Clear error path when nothing is configured
 
-    If a provider is configured but its module is missing (post-update scenario where
+    If ANY configured provider's module is missing (post-update scenario where
     `amplifier update` wiped the venv), this function will automatically reinstall
     all known provider modules without user interaction. We install ALL providers
     (not just the configured one) because bundles may include multiple providers.
+
+    The repair is gated on every configured module, not just the active one:
+    `amplifier update` and `amplifier reset` install no providers themselves, so
+    this is the single funnel for update, reset, and fresh install. Gating on the
+    active provider alone would leave a hole - if the active provider happens to
+    be installed but another configured provider is missing (a transient install
+    failure, a hand-edited settings.yaml, a single-provider `amplifier provider
+    install`), no boot would ever repair it.
+
+    Return value stays driven by the ACTIVE provider only: True means "push the
+    user into the add-a-provider prompt", which a non-active provider failing to
+    install must never trigger.
     """
     config = create_config_manager()
     provider_mgr = ProviderManager(config)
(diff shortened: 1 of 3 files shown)


bkrabach:
Confirmed — good catch, and it's a real regression. Fixed in `254c938`.

I built my own repro before applying your patch, so we have two independent reproductions of the same defect:

```python
config = _config_with_providers([
    {"module": "provider-localdev"},   # active, installed, NOT in effective sources
    {"module": "provider-openai"},     # missing, IS in effective sources
])
present = {"provider-localdev"}
...
assert needs_init is False
```

- Against this branch: **fails**, `assert True is False`, with `Failed to auto-install providers after detecting missing modules. Will prompt user to add a provider.` in the captured log.
- Against `origin/main`'s `init.py`, same test: **passes**.

So your regression claim checks out in both directions — `main` returns `False` here and we were returning `True`. Traced the mechanism to confirm it's the one you identified and not a lookalike: `install_known_providers()` only iterates `get_effective_provider_sources()`, and step 1 of that function applies `sources.modules` overrides with `for module_id in list(sources.keys())` — so it only ever augments the 8 known providers. A locally editable-installed module is never a key, never installed-into, never returned. Exactly as you described.

Applied your patch as written, both terms kept:

```python
if current_provider.module_id in installed or _is_provider_module_installed(
    current_provider.module_id
):
```

Agreed on keeping both rather than replacing. `install_known_providers()` does a manual `invalidate_caches()` / `addsitedir()` / `list(distributions())` refresh precisely because subprocess installs aren't visible to the running process — a direct-only check would trade this edge case for a false init prompt on the common path. Added a comment above the condition explaining why each term is load-bearing, so it doesn't get "simplified" later.

Sibling test added next to `test_unfixable_leftover_warns_but_does_not_block`, as suggested. Written first, confirmed failing, then fixed.

**On `still_missing` (the other `in installed` usage) — checked, not affected.** I traced every branch before deciding to leave it:

```python
still_missing = [m for m in missing if m not in installed]
```

`missing` only ever contains modules that are *not* importable at check time, so the local-dev case can't reach it — a locally-installed provider is never in `missing` to begin with. For anything that is: repair either installs it (lands in `installed`) or genuinely cannot (correctly warned). No false warning is reachable, so it doesn't need the same treatment. Happy to add the symmetric check anyway if you'd rather have the belt-and-suspenders consistency — it's a warning string, not control flow, so it's your call.

Verification after the fix:

| | passed | failed |
|---|---|---|
| `main` baseline | 1166 | 15 |
| `2cf6843` | 1174 | 15 |
| `254c938` | 1175 | 15 |

Same 15 failures throughout, identical list, all pre-existing in unrelated files. `ruff check` clean on both touched files.

Pushed as a follow-up commit rather than an amend so you can diff just the delta against what you already reviewed.

bkrabach:
You're right on (2), and it means `254c938` — the fix from your first pass — was wrong. Replaced in `c9dee10`.

To be explicit about what landed when: (1) was already fixed in `254c938`. (2) is new, and it invalidates that fix, because `in installed` short-circuits — the bad term is evaluated first, so the `or` I shipped *preserves* exactly the defect you're describing. Confirmed with a test asserting `needs_init is True` when the active provider is in `installed` but unimportable: fails against `254c938`.

Verified both of your facts before changing anything:

```python
if result.returncode != 0:
    raise RuntimeError(f"Failed to install: {result.stderr}")
...
installed.append(module_id)
```

Appended on `rc == 0`, no importability check. Your (2) holds.

**And my argument for keeping the proxy was wrong on its own terms** — worse than you put it. I justified `in installed` as a fallback against an imperfect post-install cache refresh. But `install_known_providers()` runs that refresh *itself, before returning*:

```python
    if installed:
        importlib.invalidate_caches()
        for site_dir in site.getsitepackages():
            site.addsitedir(site_dir)
        if hasattr(importlib.metadata, "distributions"):
            list(importlib.metadata.distributions())
    return installed
```

So by the time `check_first_run()` evaluates the condition, the refresh has already happened. The fallback was guarding a window that doesn't exist in this call path — it could only ever admit false positives, never prevent a false negative. Your synthetic-provider experiment shows the module is visible even earlier than that; the ordering makes the question moot either way. I should have read the call path before defending the extra term.

Applied as you wrote it:

```python
if _is_provider_module_installed(current_provider.module_id):
```

Comment above it rewritten — the old one argued *for* keeping both terms, so leaving it would have been worse than no comment. It now records why the repair's self-report isn't sufficient evidence, pointing at the same principle as `is_provider_module_installed()`'s docstring.

Agreed this retires the #238 defect rather than working around it. `still_missing` left alone — `installed` is a legitimate signal there.

**Test fallout: 4 existing tests broke, and the mocks were what was wrong, not the assertions.** Each paired a constant `_is_provider_module_installed → False` with `install_known_providers` reporting success — encoding a world where installing never makes anything importable. Made them stateful instead: the repair stub adds what it installs to the set the check reads. For the end-to-end one, which drives the real `install_known_providers`, the subprocess stub does it at the actual install boundary.

Two independent checks that this wasn't assertion-tuning:

1. Across the entire test diff, exactly one `assert` line was added — the new test's — and **zero** were removed or modified.
2. With the test changes applied but `init.py` reverted to `254c938`: 1 failed, 23 passed. All four repaired tests pass under *both* implementations. Had they been tuned to the new behavior, they'd have failed there.

| | passed | failed |
|---|---|---|
| `main` | 1166 | 15 |
| `2cf6843` | 1174 | 15 |
| `254c938` | 1175 | 15 |
| `c9dee10` | 1176 | 15 |

Same 15 failures throughout, identical list, all pre-existing in unrelated files. `ruff check` clean on all touched files; repo-wide count is 14 on this branch and 14 on `main`, unchanged.

Both siblings you asked for are present: `test_locally_developed_active_provider_is_not_forced_into_init` (still passing — the direct check subsumes the case `254c938` was reaching for) and the new `test_active_reported_installed_but_not_importable_must_not_start_session`.

Three commits now, none amended, so you can diff just `254c938..c9dee10`.

PR #258, opened 2026-08-06
fix(routing): compose routing behaviors before prepare

ramparte:
Problem

The routing hook was being appended to the final bundle_config after prepare(), but its module source was not available during prepare. That produced a "source-less hook appended after prepare" failure: the hook entry existed in bundle_config but its code could not be imported/initialized because compose/load happened too late.

What I changed

- Compose routing behaviors before calling prepare(): resolve routing config early and include its behavior URI in compose_behaviors so the routing bundle and its modules are available during prepare().
- Add required_behaviors parameter to load_and_prepare_bundle(): callers can mark a subset of composed behaviors as required; failures to load/compose required behaviors now propagate as exceptions and abort prepare. Other behaviors remain optional and still only warn.

Why

- Pre-prepare composition ensures modules/hook sources are present when the bundle's prepare step mounts/inspects hook modules; otherwise a hook entry may be appended to the prepared mount_plan without any corresponding source available.
- Routing is an app-level policy whose hook must be present during prepare; treat it as a required behavior so routing composition failures fail fast rather than silently becoming a non-functional hook in the prepared bundle.

Behavior and failure semantics

- When routing_config is active, its behavior URI is added to compose_behaviors and included in required_behaviors.
- If a required behavior fails to load or compose, load_and_prepare_bundle raises the original error and aborts preparation. This avoids the prepared bundle containing hook entries with missing sources.
- Optional behaviors (notifications, etc.) retain the previous semantics: their composition/load failures remain warnings and do not abort prepare.

Tests

- New unit tests assert that a required behavior load failure propagates (test_required_behavior_load_failure_propagates) while an optional behavior failure stays ignored (test_optional_behavior_load_failure_is_still_ignored).
- Integration-level tests assert routing behaviors are composed when routing_config is present, omitted when absent, and that prepare errors propagate when routing is active.

Compatibility risk

- Low. This tightens semantics for routing-only: systems where routing composition fails will now fail fast. Other behavior semantics are unchanged. If callers depended on routing silently degrading to a no-op, they must handle the new failure mode.

Notes

- The branch was created from feat/cli-session-matrix. This PR is targeted at that feature branch (base: feat/cli-session-matrix) so the diff contains only the fix commit(s) on top of the matrix work.
- Acceptance tests were independently run: 38 passed.

PR created from ramparte:fix/compose-routing-before-prepare -> microsoft:feat/cli-session-matrix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,5 +69,6 @@ next-steps.md
 /amplifier-app-cli/
 
 # Working folders
+.next/
 ai_working/tmp
 tests/recipes/DECISIONS.md
--- a/amplifier_app_cli/lib/bundle_loader/prepare.py
+++ b/amplifier_app_cli/lib/bundle_loader/prepare.py
@@ -51,6 +51,7 @@ async def load_and_prepare_bundle(
     source_overrides: dict[str, str] | None = None,
     progress_callback: Callable[[str, str], None] | None = None,
     bundle_source_overrides: dict[str, str] | None = None,
+    required_behaviors: set[str] | None = None,
 ) -> PreparedBundle:
     """Load bundle by name or URI and prepare it for execution.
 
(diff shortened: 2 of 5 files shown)


ramparte:
Superseded by PR #261, which is the same fix rebased onto current main. PR #261 contains the same routing behaviors composition logic but incorporates ~20 later-merged commits, providing a clean foundation without conflicts. Closing in favor of #261.

PR #282, opened 2026-08-27
feat: add Amplifier configuration skill

ramparte:
## Problem

Amplifier CLI configuration spans root-session provider selection, spawned-session routing, bundle composition, and local/project/global scope overrides. Users need one reliable workflow that identifies the controlling plane and current effective configuration before changing anything.

## Solution

Package an inline, user-invocable `/amplifier-config` skill that:

- inspects effective configuration and authoritative current sources;
- separates root provider selection from spawned routing;
- proposes the smallest safe change with provenance, blast radius, and rollback;
- applies only authorized changes and verifies observed outcomes.

## Why app-cli

The skill depends on CLI-specific configuration behavior and `app-cli:cli-expert`. Packaging it with app-cli keeps its guidance version-locked to the installed CLI and its version-matched expert.

## Files changed

- `amplifier_app_cli/data/skills/amplifier-config/SKILL.md` — packaged inline configuration workflow.
- `tests/test_merge_utils.py` — package-relative discovery and skill contract coverage.
- `tests/test_skills_cli.py` — packaged slash-command discovery and invocation coverage.

## Verification

- `uv run pytest -q`: **1470 passed, 1 skipped, 13 deselected, 1 xfailed**.
- Built wheel includes `amplifier_app_cli/data/skills/amplifier-config/SKILL.md`.
- Fresh-process pilot smoke verified slash discovery and the root-vs-spawned provider distinction.
- Focused packaged-skill checks: **3 passed**.

## Breaking changes

None.

### DIFF
--- a/amplifier_app_cli/data/skills/amplifier-config/SKILL.md
+++ b/amplifier_app_cli/data/skills/amplifier-config/SKILL.md
@@ -1,0 +1,150 @@
+---
+name: amplifier-config
+description: >-
+  Explain, inspect, troubleshoot, and safely modify Amplifier CLI configuration,
+  including provider selection, settings scopes, bundles, routing, skills, source
+  overrides, and provider-module settings composed into the CLI. Use when someone
+  describes desired Amplifier CLI behavior and needs effective-value provenance, a
+  safe change plan, implementation, or outcome-based verification.
+user-invocable: true
+version: 0.1.0
+license: MIT
+---
+
+# Amplifier CLI Configuration Consultant
+
+Translate desired Amplifier CLI behavior into the smallest correct configuration change. This skill owns the CLI-facing workflow. Portable bundle, routing, skill, and provider systems are dependencies to consult through their owning documentation and implementation, not domains this skill independently defines. Work from the current installation and effective CLI configuration, not remembered defaults or frozen examples.
+
+## Request
+
+$ARGUMENTS
+
+Treat `$ARGUMENTS` as the explicit request when present and use the current conversation for relevant context. If the desired behavior, scope, or permission to edit is unclear, ask one focused question before proceeding.
+
+In an interactive CLI session, invoke this skill with `/amplifier-config <request>`. In headless use, send a natural-language request that identifies the skill or explicitly call `load_skill` with `skill_name: amplifier-config` and pass the request as `arguments`. Do not assume a slash command embedded in a headless prompt will be dispatched as an interactive command.
+
+## Operational contract
+
+### 1. Classify the configuration plane
+
+Before recommending anything, identify every affected plane and keep them distinct:
+
+1. **Current/main-session provider selection** — which configured provider instance and model answer the root conversation, including launch selection, configured defaults, and session pinning.
+2. **Spawned/delegated routing** — how sub-agents, delegated sessions, recipe steps, and other spawned work select providers or models through roles and preferences.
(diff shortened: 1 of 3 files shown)


bkrabach:
Thanks for this, @ramparte — the underlying signal is real: users need a repeatable way to inspect *effective* Amplifier CLI configuration (root provider selection vs. spawned-session routing vs. bundle composition vs. scope precedence) instead of guessing from memory or stale docs. That's exactly the kind of friction an inline `/amplifier-config` skill should absorb.

Per our review policy for community PRs, here's the due diligence we ran before merging:

**Accuracy check against this week's changes**

The three "landmines" we specifically checked this skill against (all merged to `main` *after* this branch was cut):
- #284 (provider wizard: `None`-default fields render "leave unset" and omit the key)
- amplifier-module-provider-openai#69 (model-gated `ConfigFields`, e.g. `prompt_cache_retention`/`text_verbosity`/`enable_reasoning_context`, plus did-you-mean on unknown keys)
- #281 (multi-instance credential reuse-or-separate flow)

None of these are hardcoded anywhere in `SKILL.md`. The skill is written at the right altitude: it classifies configuration *planes* (root selection, spawned routing, bundle composition, scope precedence, runtime session state, skill discovery, provider-module fields) and requires live inspection + delegation to `app-cli:cli-expert` for anything version-sensitive, rather than freezing today's field names or wizard behavior into the doc. That design choice is exactly what keeps a config-help skill from going stale the next time a provider module or wizard ships a change — so there was nothing to revise here. We also confirmed the scope names (`global`/`project`/`local` persisted, `session` runtime-only) and the `${VAR}` env-placeholder convention it references match current `lib/settings.py` / `provider_config_utils.py`, and that `app-cli:cli-expert` is the real, existing agent reference used elsewhere in this repo (`agents/cli-expert.md`, `behaviors/cli-expertise.yaml`).

**Wiring verdict: confirmed discoverable, not inert**

`amplifier_app_cli/data/skills/<name>/SKILL.md` is auto-appended to `tool-skills`'s config by `_ensure_default_skills_dirs()` (`runtime/config.py`) — no separate index/registration step, same mechanism that already ships `goalify`, `goal-batch`, and `ten-lane-highway`. The new `test_packaged_amplifier_config_discovery_and_invocation` test in `tests/test_skills_cli.py` exercises this end-to-end (reads the packaged dir, parses frontmatter, confirms `/amplifier-config <args>` dispatches to `load_skill`), and `test_amplifier_config_skill_contract` in `tests/test_merge_utils.py` pins the frontmatter contract (name/user-invocable/license/version, no `context`/`model_role`/`allowed-tools`, `$ARGUMENTS` and `app-cli:cli-expert` present in the body). Both pass.

**Test results**

- Full suite on the PR branch as submitted (2 commits behind `main`): 1471 passed, 1 skipped, 13 deselected, 1 xfailed.
- Rebased onto current `main` (picks up #283/#284) to get an apples-to-apples baseline: **1493 passed**, 1 skipped, 13 deselected, 1 xfailed — i.e. `main`'s 1491 + this PR's 2 new tests, zero regressions. (The 1471 vs. 1491 baseline gap before rebasing is fully explained by `tests/test_provider_config_unset_fields.py`, added by #284, which this branch predates.)

**Disposition**

Accurate and wired as submitted — no revisions needed. Since this PR comes from a fork, we can't push a rebase to your branch, so we're merging the original commit as-is via squash (repo has no branch protection issue, but the ruleset requires 1 approving review — using admin-merge at team direction to land it, since CI is 9/9 green and content review is complete). Credit stays with you as author; thanks for surfacing the gap.

PR #297, opened 2026-09-03
feat(spawner): expose a timed-out sub-session's partial output via session.partial (k64 prereq 2/2, PRODUCER)

bkrabach:
## What this is

The **PRODUCER** half of the per-delegate-timeout partial-result path.
It completes the hard prerequisite of `model_performance-k64`.

The **CONSUMER** half is already merged: **amplifier-foundation `f42f48c`**
(`model_performance-bp0`, foundation PR #353). That side makes a timed-out
delegate **return** rather than raise, and read an optional app-layer
`session.partial` capability:

```
(sub_session_id: str) -> {"text": str, "segments": int, "source": str} | None
```

Nothing in the app registered one, so `partial_available` was **false for every
real timeout**. This PR is what offers one.

k64's gate **G-D4** stops the whole eval with `PARTIAL-PATH-NOT-EXERCISED`
unless at least one timeout carries `partial_available: true`. It now can.

## The non-obvious part — do not "simplify" this back

The designed patch published the partial from an `except BaseException:` handler
around `child_session.execute()`. **Against the real merged consumer that still
reports `partial_available: false`, and it does so silently — both halves' unit
tests pass.**

`tool-delegate`'s `_await_child_with_deadline` deliberately does **not** wait for
a child that is slow to unwind: at the deadline it cancels the child, **detaches**
it, and raises. Its handler then calls `session.partial` immediately — before the
cancelled child has been scheduled to run any handler of its own. Observed in one
run:

```
WARNING tool_delegate    Agent 'explorer' timed out after 1s … No partial output could be recovered.
WARNING session_spawner  Sub-session …_explorer did not complete; preserved 2 partial text segment(s), 42 chars
```

The producer preserved the work one line **after** the consumer gave up on it.

So the record is **published eagerly and updated in place**, readable at any
instant, with no dependence on cancellation ordering — and removed when the
sub-session completes normally. `test_partial_is_readable_before_the_child_has_unwound`
pins that in-repo, without needing foundation on the path.

## Evidence

All under `docs/lanes/9w0-delegate-timeout-partial-producer/evidence/`.

**Fail-before** — same cross-repo round trip, parent app-cli × real foundation `f42f48c`:

```
E  AssertionError: session.partial produced nothing -- k64's G-D4 would record PARTIAL-PATH-NOT-EXERCISED here
E  assert False is True
WARNING tool_delegate  Agent 'explorer' timed out after 1s … No partial output could be recovered.
2 failed in 3.29s
```

**Counter-evidence** — the originally designed shape (publish only from the
child's unwind), everything else identical, still `false`:

```
E  assert False is True
WARNING tool_delegate    … No partial output could be recovered.
WARNING session_spawner  … preserved 2 partial text segment(s), 42 chars      <-- too late
2 failed in 3.30s
```

**Pass-after** — patched app-cli × real foundation `f42f48c`, overlaid copies
(`cp -rL`), neither repo mutated: **2 passed**.

**G-D4 reached** — same probe, parent vs patched:

| key | parent | patched |
|---|---|---|
| `session.partial` registered | `False` | `True` |
| `partial_available` | `false` | **`true`** |
| `partial_response` | `null` | `"anchor A1 confirmed. anchor A2 confirmed. "` |
| `partial_segments` | `0` | `2` |
| `partial_source` | `"none"` | `"spawn-accumulator"` |
| `partial_chars_total` | `0` | `42` |
| `ToolResult.success` | `False` | `False` |
| `status` | `timeout` | `timeout` |

**Normal completions byte-identical** — shown, not asserted. Same probe, same
fixed sub_session_id, only `amplifier_app_cli` differs:

```
$ diff <(probe normal @parent) <(probe normal @patched)
(no differences)
ef8c86fdaf3b68e26e02e86e6a1bd59e0b638e4eb9b36f8f0a7f9f0138a35700  normal-parent.json
ef8c86fdaf3b68e26e02e86e6a1bd59e0b638e4eb9b36f8f0a7f9f0138a35700  normal-patched.json
```

The returned dict is exactly `{output, session_id, status, turn_count, metadata}`
— no new key, no shape change — and the registry is cleared on the success path.

**Suites**

| | parent `ab47608` | this branch |
|---|---|---|
| `tests/` | 1647 passed, 1 skipped, 1 xfailed | **1659 passed**, 1 skipped, 1 xfailed |
| named suites (spawner/runner/issue-233) | 73 passed, 0 failed | 73 passed, 0 failed |
| `ruff check` | 14 pre-existing errors | same 14, none in touched files |

**A correction to the item's premise, stated rather than manufactured:** the two
`TestSpawnEnrichment::test_{spawn,resume}_result_includes_status_and_turn_count`
failures reported at `f16375fc` do **not** exist at `ab47608` — both **PASS**, on
the parent and on this branch. The baseline improved between those commits.

## Ordering

The consumer half is **already merged** (foundation `f42f48c`), so the ordering
constraint from the design (`DESIGN.md §1e` — enabling the producer before the
consumer is the data-loss configuration) is satisfied. This PR is safe to merge
now.

`settings.timeout` still defaults to `None`; this PR ships mechanism only and
does not enable, sweep, or bank a timeout value.

## Scope

`amplifier_app_cli/session_spawner.py`, `amplifier_app_cli/session_runner.py`,
one new test file, and lane artifacts under `docs/lanes/`. No other repo touched.

Full write-up: `docs/lanes/9w0-delegate-timeout-partial-producer/DONE-NOTE.md`.

### DIFF
--- a/amplifier_app_cli/session_runner.py
+++ b/amplifier_app_cli/session_runner.py
@@ -512,10 +512,14 @@ def register_session_spawning(session: AmplifierSession) -> None:
     The capabilities registered:
     - session.spawn: Create new agent sub-session
     - session.resume: Resume existing sub-session
+    - session.partial: Recover a sub-session's preserved partial output after it
+      was cancelled or timed out (tool-delegate reads this to return an
+      incomplete-with-partial result instead of discarding the work)
 
     Args:
         session: The AmplifierSession to register capabilities on
     """
+    from .session_spawner import get_partial_output
     from .session_spawner import resume_sub_session
     from .session_spawner import spawn_sub_session
 
@@ -569,6 +573,7 @@ def register_session_spawning(session: AmplifierSession) -> None:
 
     session.coordinator.register_capability("session.spawn", spawn_capability)
     session.coordinator.register_capability("session.resume", resume_capability)
+    session.coordinator.register_capability("session.partial", get_partial_output)
 
 
 # =============================================================================
--- a/amplifier_app_cli/session_spawner.py
+++ b/amplifier_app_cli/session_spawner.py
@@ -21,6 +21,159 @@ from .agent_config import merge_configs
 logger = logging.getLogger(__name__)
 
 
+# =============================================================================
+# Partial-output preservation for delegates that never finish
+# =============================================================================
+# When tool-delegate's wall-clock timeout fires it cancels the spawn coroutine.
+# `child_session.execute()` raises CancelledError, the post-run block below is
+# skipped, and everything the agent produced is discarded -- turning a hang into
+# data loss. This registry keeps the agent's own assistant text so the delegate
+# tool can hand the caller an INCOMPLETE-with-partial result instead of an empty
+# failure.
+#
+# This is the PRODUCER half of the contract whose CONSUMER shipped in
+# amplifier-foundation f42f48c (tool-delegate). That side reads an optional
+# `session.partial` capability:
+#
+#     (sub_session_id: str) -> {"text": str, "segments": int, "source": str} | None
+#
+# and degrades to `partial_available: false` when it is absent, returns None, or
+# raises. Nothing here may raise into the timeout path.
+#
+# NOTE: the transcript is separately checkpointed mid-run (see
+# _install_transcript_checkpoint below) so a timed-out sub-session stays
+# RESUMABLE. That is a different property from this one: the checkpoint makes
+# the work reachable by a later resume, this registry makes it readable by the
+# delegate call that timed out, at the moment it gives up.
+#
+# WHY THE RECORD IS PUBLISHED EAGERLY (and not only on the cancellation path)
+#
+# The obvious shape -- accumulate privately, publish from an `except
+# BaseException:` around execute() -- DOES NOT WORK against the consumer that
+# actually shipped, and fails silently rather than loudly. tool-delegate's
+# `_await_child_with_deadline` deliberately does NOT wait for a child that is
+# slow to unwind: at the deadline it calls `child_task.cancel()`, DETACHES the
+# task, and raises immediately. Its `except _DelegateTimeoutExpired:` handler
+# then calls `session.partial` straight away -- while the cancelled child task
+# has not yet been scheduled to run its own exception handlers. A partial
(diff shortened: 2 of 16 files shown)


bkrabach:
## Manager verification — FIX, and this is the commit that unblocks k64's $60 eval

Head `7f749c0`, base `ab47608` = current `origin/main`. CI green on all 9 jobs.

**Full suite on the head: `1659 passed, 1 skipped, 1 xfailed`.**

### The gate that actually mattered — reproduced by me, not read off the evidence file

k64's **G-D4** stops the whole eval with `PARTIAL-PATH-NOT-EXERCISED` unless at least one timeout carries `partial_available: true`. I ran this lane's cross-repo round-trip against a **fresh clone of merged `amplifier-foundation`** (the real consumer half, `#353`), with this branch's producer on `PYTHONPATH` and neither repo mutated:

```
2 passed
```

and the lane's own captured contrast holds on the same wiring:

| | parent `ab47608` | this branch |
|---|---|---|
| `session.partial` registered | **False** | **True** |
| `partial_available` | `false` | **`true`** |
| `partial_chars_total` | 0 | 42 (2 segments, `spawn-accumulator`) |

So the producer/consumer pair is closed end to end. **`model_performance-k64` is no longer prerequisite-blocked.**

**Normal completions byte-identical** — sha256 match, diff empty. Shown, not asserted.

### The correction I want on the record, because it is the honest half

The goal told this lane to expect two named pre-existing failures (37n's `71 passed / 2 failed` at `f16375fc`) and to reproduce them on the parent. **They do not exist at `ab47608`.** I checked independently:

```
tests/test_session_spawner.py + test_session_runner.py + test_session_spawner_issue_233.py
  -> 73 passed, 0 failed   (parent)
```

The lane reported that as a correction rather than manufacturing the failures the goal predicted. That is exactly right — a stale expectation in a goal is not licence to produce matching output, and 37n's baseline had simply moved on.

Fail-before for the new test file is collection-level on the parent (new symbols), which is weak on its own — the **behavioural** parent/patched contrast above is the real evidence, same pattern as `#353`.

FIX with the decisive gate reproduced → merging. Squash + `--admin` per the base-branch policy.

PR #302, opened 2026-09-04
fix: prevent duplicate bundle registrations

bkrabach:
## What changed

- Store app bundles solely under `bundle.app` instead of also creating a selectable `bundle.added` registration.
- Make repeated additions of the same exact URI idempotent, and reject registering an exact URI under a different normal-bundle alias.
- Keep app-only sources updateable by checking and loading their exact URIs during global updates.
- Keep per-bundle registry lookup failures from aborting the global update status scan.
- Document the intentional root-only bundle load so cached modules are not refreshed redundantly.

## Why

App bundle additions were creating duplicate settings entries and app-only sources could not participate reliably in global updates. This keeps composition sources separate from selectable registrations while preserving update behavior and avoiding ambiguous URI aliases.

## Verification

- `uv run ruff check amplifier_app_cli/commands/bundle.py amplifier_app_cli/commands/update.py tests/test_bundle_commands.py`
- `uv run ruff format --check amplifier_app_cli/commands/bundle.py amplifier_app_cli/commands/update.py tests/test_bundle_commands.py`
- `uv run pytest tests/test_bundle_commands.py tests/test_get_bundle_repo_info.py tests/lib/bundle_loader/test_discovery.py tests/test_update_executor.py tests/test_windows_update_reset.py -q` (53 passed)
- `uv run pytest -q` (1693 passed, 1 skipped, 13 deselected, 1 xfailed)
- DTU validation: installed the uncommitted snapshot in an isolated DTU; app behavior add wrote only `bundle.app`; repeat remained one entry after fetching again; a duplicate normal URI alias was rejected; `amplifier update --check-only` enumerated the app-only URI. DTU `amplifier-app-cli-bundle-add-verify` and its dedicated Gitea instance were explicitly destroyed afterward.

## Breaking changes

- Newly added app bundles are now composition-only under `bundle.app` and are no longer selectable through `amplifier bundle use`; use a normal non-`--app` registration if selectability is needed.
- Registering an exact URI under a second normal bundle alias is now refused; re-use the existing name or update that name's URI.

### DIFF
--- a/amplifier_app_cli/commands/bundle.py
+++ b/amplifier_app_cli/commands/bundle.py
@@ -952,8 +952,6 @@ def bundle_add(uri: str, name_override: str | None, app: bool):
     # Use override name if provided, otherwise use name from metadata
     name = name_override or bundle_name
 
-    # All bundles are now stored in settings.yaml under bundle.added
-    # App bundles additionally go in bundle.app for composition policy
     app_settings = AppSettings()
 
     if app:
@@ -967,9 +965,8 @@ def bundle_add(uri: str, name_override: str | None, app: bool):
             console.print(f"  URI: {uri}")
             return
 
-        # Add to bundle.app (composition policy) AND bundle.added (for updates)
+        # App bundles are composition sources, not selectable registrations.
         app_settings.add_app_bundle(uri)
-        app_settings.add_bundle(name, uri)
         console.print(f"[green]✓ Added app bundle '{name}'[/green]")
         console.print(f"  URI: {uri}")
         if bundle_version:
@@ -981,6 +978,22 @@ def bundle_add(uri: str, name_override: str | None, app: bool):
     else:
         # Add to bundle.added in settings.yaml
         existing = app_settings.get_added_bundles()
+        existing_alias = next(
+            (
+                existing_name
+                for existing_name, existing_uri in existing.items()
+                if existing_name != name and existing_uri == uri
+            ),
+            None,
+        )
+        if existing_alias:
+            console.print(
+                "[yellow]Warning:[/yellow] Bundle URI already registered "
+                f"as '{existing_alias}'"
+            )
+            console.print(f"  URI: {uri}")
+            return
+
         if name in existing:
             console.print(
                 f"[yellow]Warning:[/yellow] Bundle '{name}' already registered"
@@ -1043,8 +1056,6 @@ def bundle_remove(name: str, app: bool):
         # Check if name is a URI directly in the list
         if name in app_bundles:
             app_settings.remove_app_bundle(name)
-            # Also remove from bundle.added (keeps settings in sync)
-            app_settings.remove_added_bundle(name)
             console.print("[green]✓ Removed app bundle[/green]")
             console.print(f"  URI: {name}")
             return
@@ -1058,8 +1069,6 @@ def bundle_remove(name: str, app: bool):
 
         if matching_uri:
             app_settings.remove_app_bundle(matching_uri)
-            # Also remove from bundle.added (keeps settings in sync)
-            app_settings.remove_added_bundle(name)
             console.print(f"[green]✓ Removed app bundle '{name}'[/green]")
             console.print(f"  URI: {matching_uri}")
         else:
--- a/amplifier_app_cli/commands/bundle.py
+++ b/amplifier_app_cli/commands/bundle.py
@@ -987,12 +987,20 @@ def bundle_add(uri: str, name_override: str | None, app: bool):
             None,
         )
         if existing_alias:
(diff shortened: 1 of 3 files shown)


bkrabach:
## Review: approve, with two follow-ups for the author

Reviewed by pulling the branch, reading the diff against `main`, tracing both update legs into `amplifier-foundation`, and exercising the real CLI against an isolated `HOME`.

**Core change is correct.** I reproduced the bug it fixes: on `main`, `bundle add <uri> --app` writes to **both** `bundle.app` and `bundle.added`. On this branch it writes only `bundle.app`. Full suite green locally (`1690 passed` before my commit, `1692` after), all 9 CI checks green.

Bonus worth calling out in the PR body: the `registry.load(name)` → `load_bundle(uri)` swap also removes a registry.json write-amplification — the old path flushed the entire in-memory map on every `amplifier update`, persisting all 10 well-known bundles whether or not they had ever been loaded. The new path writes nothing.

---

### Fixed and pushed as 8d7221f — `bundle add` alias refusal exited 0

The new alias-conflict guard printed a warning and `return`ed, so:

```
$ amplifier bundle add file:///... --name other-alias
Warning: Bundle URI already registered as 'teambundle'
$ echo $?
0
```

Nothing was written, but a scripted caller could not distinguish "registered under other-alias" from "refused". Every other failure in `bundle_add()` uses `sys.exit(1)`.

Now a refusal: red `Error:`, a remedy line, `sys.exit(1)`. The **idempotent** cases stay at 0 deliberately — re-adding the same name+URI, and re-adding an existing `--app` source, both end in the state the caller asked for, so they are no-ops, not refusals. Two new CLI-level tests pin that distinction. Verified end to end:

| case | exit |
|---|---|
| fresh add | 0 |
| same URI, different alias | **1** (was 0) |
| re-add same name+URI | 0 (unchanged) |
| re-add `--app` | 0 (unchanged) |

---

### 1. Please add a comment explaining `auto_include=False` (minor, but load-bearing)

`update.py:1268` now loads with `auto_include=False`, which narrows what `update_bundle()` refreshes. Measured on a parent+child probe:

```
old (registry.load):        tools = ['tool-own', 'tool-x']   # tool-x from the included child
new (auto_include=False):   tools = ['tool-own']
```

I initially flagged this as a regression, then traced the second leg and it is **not** one. `amplifier update` checks modules independently of bundles: `check_all_sources(include_all_cached=True)` → `_check_all_cached_modules` → `scan_cached_modules()` recursively walks all of `~/.amplifier/cache` and checks every mutable git-backed module against its remote, regardless of which bundle declares it — and explicitly filters out `module_type == "bundle"` with the comment *"handled by update_bundle"*. So the division of labour is deliberate: modules are the cache scan's job, bundle repos are the bundle leg's job. `update_bundle`'s module-refreshing was redundant for anything already cached.

What genuinely remains is a **prefetch gap, not staleness**: a module or sub-bundle added upstream to an included bundle since it was last composed locally is in neither the cache (nothing to scan) nor the uncomposed list. It gets fetched at next session start by `load_and_prepare_bundle`. Self-healing.

Ask: one comment at the call site saying *why* includes are skipped, so nobody "fixes" it back to `True` later and silently reintroduces the write-amplification.

### 2. `registry.find()` moved outside the `try` (one-line robustness nit)

In `_check_all_bundle_status`, `uri = registry.find(bundle_name)` is now outside the `try`, so a raising `find()` aborts the entire update check rather than skipping one bundle. It was inside before. Suggest moving it back in.

### 3. "Breaking changes: None" is not accurate

Two user-visible behavior changes worth naming in the PR body, both of which look intentional:

- App bundles are no longer selectable — `bundle use <app-bundle-name>` stops working for newly added ones, since they are no longer registered in `bundle.added` and `_load_user_registry()` only registers from that key (`discovery.py:171-183`). This is the whole point of the change, but it is a break.
- Registering the same URI under a second alias is now refused; it previously succeeded.

### Not blocking — noted for the record

- **No migration for existing installs.** Confirmed empirically: add `--app` on `main`, then `bundle remove <uri> --app` on this branch removes only from `bundle.app` and leaves the stale `bundle.added` entry, which stays listed and activatable. Confirmed with the author that this PR is scoped to preventing the situation going forward, not cleaning up existing state. `bundle remove <name>` (without `--app`) still cleans both, so there is a workaround.
- Update output now keys app bundles by raw URI, so the table shows a full `git+https://...#subdirectory=...` string where a bundle name used to be. Functional; the test even names one `"app target"`.
- `bundle remove <name> --app` matches by substring (`name in uri`), which fails for any bundle whose metadata name is not a substring of its URI. **Pre-existing**, not this PR.
- Repo-wide `ruff check` reports 14 errors on `main` as well; the changed files are clean and CI does not gate on ruff.

PR #306, opened 2026-09-06
fix(highway): watchdog re-arms at its runtime cap instead of dying silently (6c3y)

bkrabach:
## The bug

The supervisor that catches stalled lanes could stop itself, and nothing caught **that**. Measured in the batch that filed it (`model_performance-6c3y`): the watchdog hit its 12h `MAX_HOURS` cap at `2026-09-03T14:14:05Z` and the highway sat idle **~34 hours** with four lanes ENDED and unverified.

## The mechanism was worse than the ticket said

The item assumed the final message never reached anyone. `watchdog.log` says otherwise — and says something more useful:

```
line     1:  2026-09-02T14:04:27Z watchdog start ... max=12h    (generation 1)
line 33456:  2026-09-03T02:10:29Z watchdog start ... max=12h    (generation 2)
line 50060:  2026-09-03T14:14:05Z WAKE: watchdog max runtime (12h) reached - restart me...
line 50274:  2026-09-06T00:26:06Z watchdog start ...            (~34h later)
```

Between those last two lines there is **neither `wake delivered (resume ok)` nor `exit: deadline reached`** — the watchdog never exited at its cap; it was killed. The wake **was** delivered (the resumed manager's own reasoning is in the log: *"restart the watchdog before doing anything else"*), and the manager then ran `tmux kill-session -t hw-watchdog__<batch>` **while running inside that session**, as a child of `wake()`'s synchronous `amplifier run --resume`. It killed itself mid-restart, before the line that starts the replacement.

So: delivery was never the gap. *Where the responder was standing* was the gap — and a restart that depends on someone answering a message is one participant away from failing.

## The change

`highway_watchdog.sh`
- **`CAP RE-ARM GUARD`** — at the cap, `exec "$0" "$@"` (generation counter bumped) **while the batch is open**: `BATCH_DIR/lanes/` exists AND `.manager-heartbeat` is younger than `HIGHWAY_ABANDON_MAX` (default 6h). Costs the manager nothing — no wake, no `wake-needed` entry.
- **The cap's purpose is preserved, not traded away.** Abandoned batch → still winds down. `MAX_HOURS` is neither removed nor raised. Residual bound, stated in the docs: an orphan can outlive its batch by at most one cap period, never indefinitely.
- **`shopt -s execfail`** — a failed `exec` kills a non-interactive bash outright, silently converting the re-arm back into the stop this fixes. It now falls back to `exec "${BASH:-bash}" "$0" "$@"`. *Found by the test, not by reading the manual.*
- **`final_notice()`** on wind-down: writes `wake-needed` unconditionally (never eaten by `WAKE_GAP` — the one message with no next poll behind it), then dispatches `amplifier run --resume` **detached** (`setsid`, `nohup` fallback) and exits. The prompt names the trap: *"The old watchdog has ALREADY exited: do NOT run `tmux kill-session …` first."*
- **`SUPERVISION HEARTBEAT`** — `.watchdog-heartbeat` touched every poll.

`highway_status.sh` — reports `watchdog_hb_age` (JSON, additive) and `SUPERVISION LAPSED <n>s ago` when DEAD, so a lapse from **any** cause (cap, crash, kill) is read off an instrument the manager already runs first on every wake, rather than inferred from a wake that never came.

`SKILL.md` — both branches, both env knobs, the new state file, and the do-NOT-`kill-session` rule placed in Phase 6 step 1 where a manager actually hits it. Docs ship in the same change (the `2nz` lesson), plus a proposed drift-check patch under the lane artifact root covering the **docs half** too.

## Is a manager-side inverse check still needed? No.

Rejected as a *distinct* mechanism: `highway_status.sh` is already the manager's mandatory first action on every wake and already printed `watchdog=DEAD`; a second check adds a thing to remember and no new signal. And it would not have caught this outage — the manager was idle *because* no wake arrived, so anything that runs "next cycle" is dead code when there is no next cycle. The cheap half is kept (`.watchdog-heartbeat` + `watchdog_hb_age`), which is what makes "supervision lapsed, and for how long" a measurement instead of an inference.

## Tests — 13 new, subprocess, real outcomes

`tests/test_ten_lane_highway_watchdog.py`, matching the standard `2nz` set (run the real script; prove "nothing ran" with an observable sentinel — here a PATH-stubbed `amplifier` that records argv):

- re-arms while open (asserts `generation=3` from **one** spawned process), and re-arms with the **exec bit stripped**
- winds down with no lanes dir; winds down on an ancient heartbeat; re-arms again when `HIGHWAY_ABANDON_MAX` is widened (proves the wind-down is the window's doing, not an unrelated refusal)
- death notice dispatched detached, names the `kill-session` trap, and **survives a 99999s `WAKE_GAP`** while an ordinary wake in the same run is provably suppressed
- `.watchdog-heartbeat` refreshed per poll; status reports the lapse duration, and reports `-1`/"unknown" rather than "0s ago" when there is no heartbeat at all
- drift-check markers stay greppable (`CAP RE-ARM GUARD`, `SUPERVISION HEARTBEAT`)

Tests never touch the default `hw` tmux socket or a live batch.

## Full suite

```
$ uv run pytest -q
1740 passed, 1 skipped, 13 deselected, 1 xfailed in 27.16s
```

Baseline re-measured in a clean worktree at `395fa68`: **1727 passed**, 1 skipped, 13 deselected, 1 xfailed. Delta **+13** — exactly the new tests, zero regressions. (The goal file quotes 1682 for that commit; that figure appears stale, so the re-measured baseline is what the delta is taken against.)

## Spend

**$0.00** against a $0.00 authority (`0 runs x 0 arms x $0 / 1.00 = $0.00`). Pure shell/source change; no API calls, no DTUs, no infrastructure ledgered.

Lane note: `docs/lanes/6c3y-watchdog-selfrestart/DONE-NOTE.md`

### DIFF
--- a/amplifier_app_cli/data/skills/ten-lane-highway/SKILL.md
+++ b/amplifier_app_cli/data/skills/ten-lane-highway/SKILL.md
@@ -88,7 +88,7 @@ repo** (a real run left `.amplifier/bin/` behind as untracked pollution).
 | `highway_status.sh BATCH_DIR WIDTH READY` | ONE call reports every lane + watchdog liveness and **computes DEFICIT by code** | "Keep lanes full" stopped being prose the day a run sat at 1 lane with work for 10 |
 | `launch_lane.sh BATCH_DIR LANE REPO GOAL [BASE_REF]` | Worktree + branch + tmux + `/goal` session, idempotent; the ONLY writer of `manifest.tsv` | Hand-written manifests diverged on column count and broke a real batch |
 | `verify_lane.sh BATCH_DIR LANE` | Git-facts probe for one landed lane (DONE.json, ahead-count, three-dot diffstat, uncommitted work) | "Ground truth from git and the filesystem, not from what any session said about itself" |
-| `highway_watchdog.sh BATCH_DIR WIDTH SESSION_ID [INTERVAL] [MAX_HOURS]` | Detached tmux loop that re-wakes THIS session (`amplifier run --resume`) on lane-end / under-width / stale heartbeat | The highway once froze overnight because the manager stopped monitoring the moment it reported status |
+| `highway_watchdog.sh BATCH_DIR WIDTH SESSION_ID [INTERVAL] [MAX_HOURS]` | Detached tmux loop that re-wakes THIS session (`amplifier run --resume`) on lane-end / under-width / stale heartbeat; **re-arms itself at its runtime cap while the batch is open** | The highway once froze overnight because the manager stopped monitoring the moment it reported status — then froze ~34h more when the watchdog hit its own cap |
 | `infra_ledger.sh BATCH_DIR add TYPE ID DESTROY_CMD...` / `infra_ledger.sh BATCH_DIR sweep --all-owners` | Records any infrastructure a lane OR the manager stands up (DTU, gitea instance, container, service, background process) into `infra.tsv` at creation, each with its teardown command; `sweep` runs those commands and exits non-zero until nothing is left standing | A run closed with a DTU and a gitea container still live — nothing the highway stands up should outlive it (Rule 14) |
 
 **`sweep` is the MANAGER's batch-close verb, never a lane's.** It runs EVERY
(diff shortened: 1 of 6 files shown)


bkrabach:
## Manager verification — the diagnosis is excellent and CORRECTS MINE. Held as draft on one gate: **macOS CI is red on this PR's own new tests.**

### First, the part that is better than what I filed

I filed `model_performance-6c3y` claiming the watchdog *"exited at its own cap and left one trace in a file nobody reads."* **That was wrong, and this lane proved it from the log.** Verified against `watchdog.log` myself:

```
50060: 2026-09-03T14:14:05Z WAKE: watchdog max runtime (12h) reached ...
50271:   echo "=== killing expired watchdog (12h04m) ==="; tmux -L hw kill-session -t "$WD" ...
50274: 2026-09-06T00:26:06Z watchdog start ...        <- the ~34h gap
```

The wake **was delivered**. The resumed manager — running *inside the watchdog's own tmux session*, as a child of `wake()`'s synchronous `amplifier run --resume` — then ran `tmux kill-session` on that watchdog and **killed itself mid-restart**, before reaching the line that starts the replacement. As the note puts it: *"Delivery was never the gap; where the responder was standing was the gap."*

That is a materially better root cause than the one in the item, and it means my proposed "announce your own death via the wake path" fix would not have helped — the announcement already worked.

The test suite matches `2nz`'s standard (real script via subprocess, PATH-stubbed `amplifier` recording argv as an observable sentinel, isolated on socket `hw-test-6c3y` so the live batch is never touched), and `shopt -s execfail` was found *by a test*, not by reading the manual — exactly the right way round.

### The gate that holds it: macOS is red, on tests this PR adds

```
pytest (macos-latest, py3.11): FAILURE      pytest (macos-latest, py3.12): FAILURE
pytest (ubuntu-latest, 3.11/3.12): SUCCESS  pytest (windows-latest, 3.11/3.12): SUCCESS
```

`tests/test_ten_lane_highway_watchdog.py` does not exist on `main`, so these are **not** pre-existing failures — they are this PR's:

```
tests/test_ten_lane_highway_watchdog.py:196  AssertionError: watchdog did not re-arm past its cap
tests/test_ten_lane_highway_watchdog.py:226  AssertionError: non-executable watchdog did not re-arm
                                             AssertionError: ancient heartbeat: watchdog did not wind down at its cap
```

So on macOS the **re-arm does not happen and the wind-down does not happen** — i.e. both branches of the deliverable are unverified there.

### Why I am not merging it anyway, even though this highway runs on Linux

The DONE-NOTE already says (line 131) *"The script was already GNU/Linux-only (`stat -c %Y`)"* — which is very likely the real explanation, and if so it is a **pre-existing** platform constraint this PR merely made visible by being the first to test the script.

**That is a fine answer. It just has to be the stated one.** Right now the PR asserts both branches are tested while CI says they fail on a supported platform, and a red CI merged on a verbal explanation is exactly how a real regression gets waved through later.

**Smallest path to merge — pick either:**
1. **Skip with a reason**, not silently: `@pytest.mark.skipif(sys.platform != "linux", reason="highway_watchdog.sh is GNU/Linux-only: stat -c %Y, ...")`. This makes the existing constraint explicit and turns CI green honestly.
2. **Make them portable**, if the failure is only the `stat` flavour (`stat -f %m` on BSD/macOS).

Either way, please say in the DONE-NOTE whether the *script* is intended to be Linux-only — if it is, that is worth stating once, loudly, rather than being rediscovered by the next lane.

Everything else here is merge-ready; this is one gate, and the fix is small.

bkrabach:
## macOS gate: cause found and remedied; CI is not re-triggering on this branch

I pushed the same fix that cleared #313 (merged `4420e14`) to this branch as `b8e6c1b`.

**The cause was never the tests or the fix** — it is a platform limitation the script documents about itself:

```
highway_status.sh:53   # (stat -c %Y below is GNU/Linux; adjust for macOS if this ever travels.)
```

`stat -c %Y` is a GNU coreutils flag with no BSD/macOS equivalent. The existing guard caught only Windows and missing-bash — but **macOS has bash**, so it never fired, and the suite ran against a script that cannot work there. Hence `live == 0`: the liveness sensor was never going to report anything on that platform.

The remedy is the one applied to #310 (Windows/bash) and to #313: **skip off GNU/Linux with the reason stated**, never waived silently. Linux CI — where these scripts actually run — is unchanged.

**Current state, reported honestly rather than left as a surprise:** after pushing `b8e6c1b` and marking this PR ready, **zero pytest jobs have been queued** across ~12 minutes of polling. #313 queued and went green immediately under identical treatment, so this is specific to this branch/workflow, not to the change. I am not merging on an unverified CI state — the same standard that held this PR in the first place.

The work itself remains verified from my earlier review: the root cause (the resumed manager killed the watchdog session it was standing in, then died mid-restart before reaching the line that starts the replacement) is a **better diagnosis than the item I filed**, the re-arm/wind-down branches are both tested, and the cap's anti-orphan purpose is preserved. It needs CI to run, not more work.

PR #311, opened 2026-09-06
fix(routing): list a live matrix the runtime loads but the CLI required a `name:` for (teel)

bkrabach:
## The defect

`amplifier routing list|show` told an operator the **opposite of the truth** about a matrix that was actively routing every agent in the session. Measured on a live host by lane `otr`, with `routing.matrix = anthropic-knob-consistent` in effect and the session resolving **13 agents** through it:

```
Matrix 'anthropic-knob-consistent' not found
0 active, 9 disabled
```

This is the tool a human reaches for *precisely when routing is misbehaving* — the moment "not found / 0 active" is most expensive. It invites exactly the wrong next action (re-point config, re-copy the file, disable the preset) against a system that was working.

## Root cause — and it is not the one the report inferred

The report attributed it to the CLI keying on the YAML's internal `name:` while the runtime keys on the **file stem**. **PR #294 (`8c4ad7a`) already fixed that half** — rows have been keyed by stem since 2026-09-02. What #294 left behind is the **existence gate**:

```python
if data and "name" in data:
    matrices[stem] = (data, path)
```

The runtime never reads that field. Checked, not assumed, against the cached bundle at `972b0ce`:

| Runtime function | What it actually requires |
|---|---|
| `resolve_matrix_source(name, …)` | builds `f"{name}.yaml"` — the **stem**, first hit in `[*custom_routing_dirs, bundle routing/]` |
| `load_matrix(path)` | only that the file parse to a **mapping** |
| `validate_matrix(matrix)` | `general` + `fast` roles, per-role `description`/`candidates` |

`grep '"name"' matrix_loader.py resolver_class.py` returns **nothing**. Its own docstring (`matrix_loader.py:41-43`) says the resolved name is the file stem and *"not necessarily the `name:` field declared inside the YAML"*.

So a valid, loadable, actively-routing matrix that simply carried no `name:` was **dropped from the CLI entirely** — no row in `list` (contributing to neither the active nor the disabled count, hence `0 active`) and `not found` in `show`. That requirement was the CLI's own invention.

## Before / after — same tree, same commands

**At the parent commit `3bb0104`:**
```
── routing matrices (0 active, 9 disabled) ──
  [off]    anthropic  (available)  ← disabled
  …
$ amplifier routing show
Matrix 'anthropic-knob-consistent' not found. Available: anthropic, balanced, …
```

**On this branch:**
```
── routing matrices (1 active, 9 disabled) ──
  [on]  → anthropic-knob-consistent  (active)

$ amplifier routing show
Matrix: anthropic-knob-consistent
  Knob-consistent anthropic routing.
```

## The change — reporting path only, three edits

1. **The existence gate becomes the loader's own rule.** `if data and "name" in data` → `if isinstance(data, dict) and data`. A matrix exists for the CLI exactly when the runtime can load it. An empty, non-mapping or unparseable file still gets **no** row — the runtime cannot load it either, and widening past that would be the same class of lie in the other direction.
2. **`_disagreeing_name()`** distinguishes a `name:` that is **absent** from one that **disagrees**. Absence draws no warning: the runtime never reads the field, so warning about it would send an operator hunting a problem that does not exist — the same wasted trip the "not found" message caused. A genuine mismatch still warns exactly as #294 made it.
3. **`show` headings name the stem the runtime resolved**, not `data["name"]` — which printed a string the runtime never reads, and rendered as the actively misleading `Matrix: unknown` for a file with no `name:`. `routing edit`'s unsaved working copy has no stem to offer and keeps its previous heading.

The report offered three candidate fixes. This is **(1)** — one source of truth, both keying *and* existence derived from the runtime's rule. **(3)** (validate and warn) was already present and is kept, narrowed to fire on real disagreement only. **(2)** (report both identities co-equally) was rejected: it implies the declared name has standing in resolution, and it has none.

## Runtime selection is UNCHANGED

This is a reporting fix. `resolve_winning_paths()` still decides which **file** a stem loads; this changes only which rows survive to be shown. `TestRuntimeSelectionUnchanged` pins that the settings bytes and the per-stem winner are identical across `list`/`show`, and that reporting commands never write settings.

## The cross-repo boundary — reported, not crossed

The runtime identity lives in `matrix_loader.py` in **amplifier-bundle-routing-matrix**. **Nothing there was edited.**

**What a matching change in that repo would need to be: none.** The runtime's rule was already correct and already documented; the CLI was the half that disagreed, and the CLI is the half that moved.

`TestCliAndRuntimeCannotDiverge` pins the CLI side by **asking that function directly** — obtained through production's own `load_resolve_matrix_source()` — and asserting the CLI's row set equals the set the runtime resolves. It cannot be satisfied by teaching the test the same wrong answer as the code, and a non-vacuity case pins that an unresolvable stem is absent from both sides.

Two optional suggestions for that repo's owner, neither required here: export the identity rule as a function so the CLI can import it rather than mirror it; and warn at load time on a `name:`/stem disagreement (the runtime is still silent, so a user who never runs `routing list` never learns the field is decorative).

## No regression for the common case — byte-identical

Nine invocations (`list`, `list --format json`, bare `show`, `show <stem>`, `--detailed`, `--compact`, a custom matrix, the not-found path, `use`) against an agreeing tree, captured at the parent commit and on this branch by stash-compare at a fixed width:

```
parent: 6497 bytes   branch: 6497 bytes
diff -u before after  ->  BYTE-IDENTICAL
```

Re-verified after `ruff format`. Held in-test by `TestAgreeingTreeUnchanged`, which pins the agreeing-tree listing as an exact golden literal. Titling by stem is a no-op *exactly* when stem and `name:` agree, which is why the common case cannot move.

## Tests

`tests/test_routing_cli_runtime_identity.py`, 28 tests:

- **fail-before at `3bb0104`: 12 failed, 16 passed** — the 16 that pass are the stock-tree and non-vacuity gates, which *should* pass at the parent; the 12 are the defect.
- **after: 28 passed.**

## Full suite

```
parent 3bb0104 : 2 failed, 1776 passed, 1 skipped, 13 deselected, 1 xfailed
this branch    : 2 failed, 1804 passed, 1 skipped, 13 deselected, 1 xfailed
```

+28 = exactly the new file. No test changed status in either direction.

**The 2 failures are pre-existing and unrelated**, identical at the parent commit. Root-caused rather than waved past: the highway exports `HIGHWAY_TMUX_SOCKET` into every lane session, and `test_ten_lane_highway_socket_default.py::_derived_socket` shells out with no `env=`, so the `${HIGHWAY_TMUX_SOCKET:-…}` default branch those two tests exist to exercise is never taken:

```
$ echo $HIGHWAY_TMUX_SOCKET                                       -> hw-model-performance
$ uv run pytest tests/test_ten_lane_highway_socket_default.py -q  -> 2 failed, 9 passed
$ env -u HIGHWAY_TMUX_SOCKET uv run pytest … -q                   -> 11 passed
```

CI does not see it (no such variable on GitHub runners), so it is green there and red only for the lanes that run it. Filed as `model_performance-etuz`, linked discovered-from this item; not fixed here, as that file is outside this lane's owned paths.

## Spend

**$0.00** against a $0.00 authority (`0 runs × 0 arms × $0.00 / 1.00`). Pure code-and-test change: no API calls, no containers, no DTU, no eval runs, no infrastructure created.

---

Item: `model_performance-teel`. Lane note with full evidence: `docs/lanes/teel-routing-cli-identity/DONE-NOTE.md`.

### DIFF
--- a/amplifier_app_cli/commands/routing.py
+++ b/amplifier_app_cli/commands/routing.py
@@ -182,10 +182,29 @@ def _load_all_matrices_with_paths(
 
     Only the winning file is parsed into the row, so a shadowed file can never
     supply the description, ``updated:`` date or compatibility count shown for
-    a matrix. A stem whose winner is unparseable or carries no ``name:`` is
-    dropped entirely rather than falling back to a file the loader would not
-    read -- that keeps the old "must have ``name:`` to be listed" rule without
-    letting a loser back in through it.
+    a matrix. A stem whose winner is unparseable is dropped entirely rather
+    than falling back to a file the loader would not read.
+
+    A row survives on exactly the loader's own terms: ``load_matrix()`` accepts
+    any file that parses to a mapping (``"Matrix file must contain a YAML
+    mapping"``) and ``validate_matrix()`` asks for ``general``/``fast`` roles.
+    **Neither reads the YAML's ``name:`` field** -- as of routing-matrix
+    ``972b0ce``, ``matrix_loader.py`` and ``resolver_class.py`` contain no read
+    of it at all. This function used to require it anyway::
+
+        if data and "name" in data:
+            matrices[stem] = (data, path)
+
+    which made a matrix the runtime loads and routes through invisible here:
+    absent from ``list`` (contributing to neither the active nor the disabled
+    count) and "not found" in ``show``. Measured on a live host with
+    ``routing.matrix = anthropic-knob-consistent`` in effect and 13 agents
+    routing through it, the CLI reported ``Matrix 'anthropic-knob-consistent'
+    not found`` and ``0 active, 9 disabled`` -- the opposite of the truth, from
+    the tool an operator reaches for precisely when routing is misbehaving.
+
+    ``name:`` is CLI-facing metadata. It is reported when it disagrees with the
+    stem (see :func:`_print_name_stem_note`), and never decides existence.
 
     Args:
         matrix_files: Every discovered matrix file.
@@ -199,7 +218,10 @@ def _load_all_matrices_with_paths(
     matrices: dict[str, tuple[dict[str, Any], Path]] = {}
     for stem, path in winners.items():
         data = _load_matrix(path)
-        if data and "name" in data:
+        # The loader's rule, not ours: a non-empty YAML mapping. An empty or
+        # non-mapping file raises in load_matrix(), so it has no runtime
+        # identity to report and must not gain a row here either.
+        if isinstance(data, dict) and data:
             matrices[stem] = (data, path)
     return matrices
 
@@ -222,6 +244,21 @@ def _load_all_matrices(matrix_files: list[Path]) -> dict[str, dict[str, Any]]:
 def _declared_name(data: Mapping[str, Any]) -> str:
     """The ``name:`` field inside a matrix YAML (may differ from its stem)."""
     return str(data.get("name", ""))
+
+
+def _disagreeing_name(data: Mapping[str, Any], stem: str) -> str | None:
+    """The declared ``name:`` when it is present AND differs from *stem*.
+
+    ``None`` covers both "agrees" and "not declared at all", which are the two
+    cases an operator must not be warned about. Absence is not disagreement:
+    the runtime never reads ``name:``, so a file without one is entirely
+    ordinary, and a warning there would send someone looking for a problem
+    that does not exist -- the same wasted trip the "not found" message caused.
+    """
+    if "name" not in data:
+        return None
+    declared = _declared_name(data)
+    return declared if declared != stem else None
 
 
 def _print_matrix_not_found(
@@ -461,8 +498,8 @@ def routing_list(compact: bool, detailed: bool, fmt: str):
 
         origin = origins.get(source_path.stem)
         is_shadowing = origin is not None and origin.is_shadowed
-        declared = _declared_name(data)
-        name_disagrees = declared != name
+        declared = _disagreeing_name(data, name)
+        name_disagrees = declared is not None
 
         config_summary: dict[str, Any] = {
             "description": description,
@@ -482,7 +519,7 @@ def routing_list(compact: bool, detailed: bool, fmt: str):
         # `routing use` must write. `declared_name` is reported alongside so a
         # disagreement is visible instead of silently keying on the wrong one.
         item["matrix_file"] = _display_path(source_path)
-        if name_disagrees:
+        if declared is not None:
             mismatched.append((name, declared))
             config_summary["declared_name"] = declared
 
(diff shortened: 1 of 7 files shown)


bkrabach:
## Manager verification — the scope correction is legitimate. Both deviations adjudicated. Merging.

Head `2c3b697`, base `3bb01040` = current `origin/main`. CI green 9/9.

### Deviation 1 — the item's inferred cause was already fixed; the lane fixed the half that still reproduced. **Verified, and this is the right call.**

`8c4ad7a` (**#294**) is on main — it moved the CLI's row **keying** from the YAML's `name:` to the file stem. What it left behind, at `commands/routing.py:168` and `:202` on the parent:

```python
if data and "name" in data:
```

The **existence gate** still required `name:`. And the runtime never reads that field at all — I grepped `matrix_loader.py`: **zero** matches for `"name"`. So a valid, loadable, actively-routing matrix carrying no `name:` was dropped from the CLI entirely — absent from `list` (hence `0 active`) and `not found` in `show`. That is exactly otr's measured symptom, and it still reproduced.

**Fail-before at the parent, measured myself: `12 failed, 16 passed` — matching the marker exactly.** On the branch: `28 passed`.

This is a scope correction, not a substitution: the item inferred a cause, half of it had been fixed under it, and the lane fixed the half that still produced the observed lie rather than reporting the item as already-fixed. It says so plainly in its own deviations.

**Runtime routing is untouched — verified explicitly**, since that was the whole premise. Outside `tests/` and `docs/lanes/`, the diff touches exactly one file: `amplifier_app_cli/commands/routing.py`. Nothing in `amplifier-bundle-routing-matrix` was edited, and the DONE-NOTE states the matching change there would need to be **none** — the runtime's rule was already correct.

I also like the shape of the anti-divergence test: it obtains `resolve_matrix_source()` through production's own loader and asserts the CLI's row set **equals** what the runtime resolves — so it cannot be satisfied by teaching the test the same wrong answer as the code.

### Deviation 2 — "two pre-existing failures". **Confirmed, and I found a third that is NOT a regression.**

Full-suite failure sets, compared:

| | parent `3bb0104` | branch `2c3b697` |
|---|---|---|
| teel's own tests | **12 failed** | 0 |
| `test_ten_lane_highway_socket_default.py` (×2) | **failed** | **failed** |
| `test_truststore_wrap_bio_shim.py` (×1) | passed | **failed** |
| totals | `14 failed, 1792 passed` | `3 failed, 1803 passed` |

- The **two socket tests are genuinely pre-existing** — red at both commits. Filed by this lane as **`model_performance-etuz`**. They are **my own defect**: I wrote and merged that file in `3bb0104`, and its `_derived_socket` helper inherits an ambient `HIGHWAY_TMUX_SOCKET`, so inside any highway lane the default derivation never runs. Confirmed live: `HIGHWAY_TMUX_SOCKET='hw-model-performance'` is exported in the lane environment. Mine to fix, not this lane's.
- The **truststore failure is test-pollution, not a regression**: it **passes in isolation on both parent and branch** (`11 passed` each). Adding a new test file changes collection order and surfaces a latent ordering dependency. No behaviour change, but worth recording rather than absorbing.

So the marker's claim is accurate for the deterministic failures, and the third is an ordering artifact I verified rather than assumed.

PR #313, opened 2026-09-06
fix(ten-lane-highway): report infra-ledger rows no live lane owns (ye80)

bkrabach:
## What this is

The **second and third sub-findings** of the 2026-09-05 tmux-server incident
(`model_performance-ye80`). The headline fix — the per-batch tmux socket —
already landed in `3bb0104`; its tests were repaired in `569c9b8`. This PR
closes what was left open, both of which cost real money that day.

**Sub-finding 2 lands here as code. Sub-finding 1 ships as a lane artifact**,
because its only target lives in another repo — see "Goal defect" below.

---

## Sub-finding 2 — nothing joined lane liveness to row ownership

A tmux server restart killed three lanes at once and left **six DTU containers
RUNNING with open ledger rows**. Every component worked in isolation: the
ledger recorded all six correctly, and `highway_status.sh` reported the lanes
ENDED. **No code path read both**, so "infrastructure with nothing driving it"
(Rule 14) was invisible — the manager found the containers only by running
`incus list` by hand.

`highway_status.sh` now reports:

```
SUMMARY batch=… live=1 ended=1 … watchdog=DEAD orphan_rows=6
WARNING: orphan_rows=6 owned by: drbf-compaction-notice-ab(6)
  Open infra-ledger rows whose owning lane is NOT live - infrastructure
  with nothing driving it (Rule 14). Nothing was destroyed; reclaim each
  with: lane_teardown.sh <BATCH_DIR> <lane> teardown --yes
```

and in `HIGHWAY_JSON=1` mode: `"orphan_rows":6,"orphan_owners":"drbf-compaction-notice-ab(6)"`.

**Reporting only.** Reaping is deliberately not implemented: a false positive
that prints is a nuisance, a false positive that destroys is another `0rg`.
`test_reporting_destroys_nothing` proves inertness with a `touch <sentinel>`
destroy_cmd plus a byte-identical ledger — not inferred from an exit code.

### Replayed against the real incident data

A copy of the live batch's `manifest.tsv` / `infra.tsv` / `infra.owners.tsv`,
with drbf's six rows set back to `open` (`evidence/04-incident-replay.txt`):

```
{"batch":"replay","live":1,"ended":147,…,"orphan_rows":6,
 "orphan_owners":"drbf-compaction-notice-ab(6)"}
```

The live batch today reports `orphan_rows:0`, matching its zero open rows —
the report is not crying wolf.

### Why owner names are resolved, not string-compared

The two sides of this join genuinely disagree about what a lane is called.
Measured in the live batch's own files: `infra.owners.tsv` holds **short
work-item ids for 5 of 13 owners** (`vbs`, `127`, `3d2`, `6da`, `k64`) and
**long manifest names for the other 8**. Resolution is exact-name, else a
**unique** token-boundary prefix. An exact-match join would have false-alarmed
on ~40% of this batch's owners. Ambiguous (`l1` vs `l1-alpha`/`l1-beta`),
unresolvable, and unclaimed owners are each reported **as such**, never guessed.

### A live lane's rows are NEVER counted

The deliverable most easily faked by making everything pass. Pinned three ways:

- `test_a_live_lanes_rows_are_never_reported_as_orphaned` — live lane owns 2
  rows, dead lane owns 6, **simultaneously**; asserts `orphan_rows == 6` *and*
  the live lane's name is absent. Counting every open row would also pass the
  incident test; this is what separates a real join from a row count.
- `test_real_tmux_agrees_with_the_stub` — same scenario against a **real tmux
  server**, then kills the session mid-test and asserts the same batch flips
  1 → 2 orphans with nothing else changed.
- Harness CASE 3 — the live lanes' containers survive an exact-name teardown.

---

## Sub-finding 1 — `lane_teardown.sh`'s near-miss footgun (artifact, not code)

Measured during the recovery, with six containers still billing:

```
lane_teardown.sh <batch> drbf teardown --yes
  -> "TEARDOWN: lane 'drbf' owns no open rows -- nothing to do"    (exit 0)
```

**Root cause, traced through the code.** The rows are `val-drbf-a…c2`, claimed
to `drbf-compaction-notice-ab`. `classify()` consults **claims first**, returns
`OTHER:drbf-compaction-notice-ab`, and the **inference** rule that *would* have
matched `val-drbf-*` for lane `drbf` is never reached. Every candidate lands in
PROTECTED, `sel_ids` is empty, and the empty-selection branch prints a success
message and exits 0.

Fixed, and proven fail-before/pass-after — but **shipped as a patch**, see below.
The guard is narrow on purpose (an id inference would have selected, or an owner
name prefix-related to `$LANE` **in either direction**), exits **4**, and names
each candidate with its open-row count and the corrected command. It is *not*
"any protected row is an error": most lanes provision nothing and run teardown
anyway, and failing them because an unrelated lane holds rows would be a
regression. CASE 4 pins that.

```
FAIL-BEFORE (unpatched):  RESULT: FAIL -- 3 expectation(s) unmet
PASS-AFTER  (patched):    RESULT: PASS -- near-miss guard present, protection intact
```

The exact-name path is unchanged, reproducing the incident's own recovery
transcript: `verified-gone=6 rows-flipped=6 … protected-untouched=4`, with each
of the four live-lane containers asserted **still present** in the stub's state.

---

## One fix or two?

**One cause, two sites, and they must ship separately.** The cause is single:
nothing reconciles a lane's identity across `manifest.tsv`, `infra.owners.tsv`
and `infra.tsv`, so a lane-name lookup returns "nothing" instead of "I could
not tell". The blast radii are not single, and that decides the shape:

- **`lane_teardown.sh` DESTROYS.** Its fix must fail **closed** — refuse, exit
  non-zero, name the candidates — and stay conservative.
- **`highway_status.sh` only REPORTS.** Its fix can fail **open** — report,
  name the reason — because a false positive costs a printed line.

Merging them into one mechanism would force one policy onto the wrong side.

---

## Goal defect (reported, not absorbed)

`lane_teardown.sh` lives at
`/home/bkrabach/dev/openai-evals-team-ci/.amplifier/evaluation/tools/lane_teardown.sh`
— **a different repo, and untracked even there**:

```
$ git -C …/openai-evals-team-ci ls-files --error-unmatch \
      .amplifier/evaluation/tools/lane_teardown.sh
error: pathspec '…' did not match any file(s) known to git
```

This lane owns `amplifier-app-cli`, which ships `infra_ledger.sh`,
`highway_status.sh`, `launch_lane.sh`, `verify_lane.sh`, `highway_watchdog.sh`
— and no `lane_teardown.sh`. Per the goal's own rule, the fix is shipped as a
patch artifact and the other repo was not touched.

**The deeper defect** (lane `2nz` already flagged it): the shipped skill
**names a tool it does not ship**. `infra_ledger.sh`'s refusal message tells the
operator to run `.amplifier/evaluation/tools/lane_teardown.sh` — an evals-repo
path — and `SKILL.md` names it with no path at all. Until that tool is adopted
here, every fix to it is unversioned. Recommended as its own item; deliberately
not smuggled into this PR, since a second copy of a 409-line tool the manager
runs live from the other path is worse than the footgun.

---

## Tests

```
$ python -m pytest -q
1 failed, 1818 passed, 2 skipped, 13 deselected, 1 xfailed in 12.86s
```

The failure is `tests/test_fail_loud_bundle_activation.py::…::test_error_is_a_bundle_error`
(`assert issubclass(ModuleActivationError, BundleError)`), a foundation-contract
assertion. **Proven pre-existing** — `git stash -u`, clean HEAD:

```
1 failed, 1805 passed, 2 skipped, 13 deselected, 1 xfailed in 12.24s
```

Identical failure; the delta is exactly this PR's +13 tests.

**Two corrections to the goal's KNOWN section:** the failure it predicted
(`test_truststore_wrap_bio_shim.py::test_real_truststore_is_covered_at_cli_import`)
did **not** occur in either run — that prediction is stale — and the failure
that *does* occur is a different one the goal does not mention. A later lane
should not absorb it either.

New-file lint clean (`ruff check`, `ruff format --check`); the 6 `ruff check
tests/` errors are pre-existing unused imports in two untouched files.

## Spend

**$0.00** of the $0.00 authority. No container, no DTU, no eval, no API spend —
as the goal's arithmetic (`0 × 0 × $0 / 1.00 = $0.00`) predicted. No
infrastructure registered, claimed, or torn down; the live ledger was never
written to, and `sweep` was never run.

## Artifacts

`docs/lanes/ye80b-teardown-orphan-rows/` — `DONE-NOTE.md`,
`PROPOSED-evals-lane-teardown-near-miss.patch`,
`test_lane_teardown_near_miss.sh`, and `evidence/00…07`.

### DIFF
--- a/amplifier_app_cli/data/skills/ten-lane-highway/SKILL.md
+++ b/amplifier_app_cli/data/skills/ten-lane-highway/SKILL.md
@@ -113,6 +113,21 @@ perform that teardown. A REAL teardown failure still exits non-zero and leaves
 the row open; the already-gone signature is deliberately narrow, so the signal
 that a teardown genuinely failed is never lost.
 
+**`highway_status.sh` reports `orphan_rows=N owned by: <lane>(n)`** — open
+ledger rows whose owning lane is **not live**. On 2026-09-05 a tmux server
+restart killed three lanes at once and left six DTU containers RUNNING with
+open rows; the ledger recorded all six correctly and the status report showed
+the lanes ENDED, but **nothing joined the two**, so the manager found the
+containers only by running `incus list` by hand. A non-zero `orphan_rows` means
+infrastructure with nothing driving it (Rule 14) — reclaim each named lane with
+`lane_teardown.sh BATCH_DIR <lane> teardown --yes`. The report **destroys
+nothing**: a false positive that prints is a nuisance, a false positive that
+destroys is another 0rg. A **live** lane's rows are never counted. Owners are
+matched by exact lane name, else by a unique lane-name prefix (so a row claimed
+with the short work-item id `vbs` still resolves to lane `vbs-…`); an owner
+that matches no lane, matches ambiguously, or is unclaimed is reported as such
+rather than guessed at.
+
 State lives in `BATCH_DIR` (create one per highway, e.g. `~/dev/hw-<name>`):
 `manifest.tsv` (scripts write), `HIGHWAY.md` (you write), `goals/` (pre-composed
 goal files), `lanes/` (worktrees), `.width` (authoritative width), `infra.tsv`
--- a/amplifier_app_cli/data/skills/ten-lane-highway/scripts/highway_status.sh
+++ b/amplifier_app_cli/data/skills/ten-lane-highway/scripts/highway_status.sh
@@ -56,6 +56,8 @@ BATCH=$(printf '%s' "$(basename "$BATCH_DIR")" | tr -c 'A-Za-z0-9_-' '_')
 
 now=$(date +%s)
 live=0; ended=0; done_n=0; stalled=0; gone=0
+# Lane-name sets for the ORPHAN-ROWS join below (newline-joined, awk splits them).
+live_lanes=""; all_lanes=""
 
 [ "$JSON" = 1 ] || printf '%-20s %-6s %-9s %-6s %-5s %-14s %s\n' LANE TMUX LOG_AGE AHEAD DONE FLAG WORKTREE
 while IFS=$'\t' read -r lane wt branch base tmuxn goal log ts; do
@@ -72,11 +74,14 @@ while IFS=$'\t' read -r lane wt branch base tmuxn goal log ts; do
     wt_present=no
   fi
 
+  all_lanes="$all_lanes$lane"$'\n'
+
   flag=""
   if [ "$st" = "LIVE" ]; then
     # A live tmux-session counts as a live lane even if its worktree vanished —
     # that is an anomaly to flag, not a lane to ignore.
     live=$((live+1))
+    live_lanes="$live_lanes$lane"$'\n'
     if [ "$wt_present" = "no" ]; then flag="NO-WORKTREE"; fi
     if [ "$age" != "-" ] && [ "${age%s}" -gt "$STALL_SECS" ]; then flag="STALLED"; stalled=$((stalled+1)); fi
   else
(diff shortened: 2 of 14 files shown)


bkrabach:
## Manager verification — the work is right. **Held as draft on one gate: macOS CI is red on this PR's own new tests.**

### Both fail-befores reproduce against current main — I ran them myself

```
sub-finding 2 (liveness/ownership join), tests vs origin/main : 13 failed
                                        same tests on branch  : 13 passed   (Linux)
sub-finding 1 (near-miss), harness vs the REAL unpatched upstream lane_teardown.sh:
                                        RESULT: FAIL -- 3 expectation(s) unmet
```

The second one is the part I most wanted to see: the harness takes the script under test as an argument, so I pointed it at the actual `.amplifier/evaluation/tools/lane_teardown.sh` this batch uses and watched it fail. That is a real fail-before against the real artifact, not a mock of one.

### The root cause is better than the item I filed

I filed two sub-findings as if they were two bugs. They are **one**: nothing reconciles a lane's identity across the three files that describe it. `manifest.tsv` holds LONG names, `infra.tsv` ids are SHORT-form, and `infra.owners.tsv` holds **either** — measured **8 long names and 5 short work-item ids among 13 distinct owners in this batch**, while `lane_teardown.sh`'s own usage text documents only the short form. A claim under the long name makes `classify()` return `OTHER:<lane>` *before* the inference rule that would have matched, so the near-miss selects nothing and exits 0.

### The four deviations, adjudicated

**(a) Reporting-only, no reaping — correct, and I want it on the record why.** The goal names reporting as sufficient and reaping as riskier. The lane's own framing is the right one: *"a false positive that destroys is another 0rg."* Verified the two halves have deliberately **opposite** failure policies — `highway_status.sh` (reports) fails **OPEN** and over-reports rather than guess; `lane_teardown.sh` (destroys) fails **CLOSED** with exit 4. That asymmetry is the whole design and it is correct.

**(b) The real-tmux test exists and runs** — `test_real_tmux_agrees_with_the_stub` at line 383, guarded by `skipif(shutil.which("tmux") is None)`. It **failed at the parent and passes on the branch** on this host, so the stub genuinely cannot hide an integration break.

**(c) Explicit subprocess env — verified.** This is exactly `etuz`, which I caused: my own test inherited `HIGHWAY_TMUX_SOCKET` and was green in CI while red inside every lane. Good.

**(d) The shell harness instead of a pytest — acceptable, and it must be.** A pytest in this repo cannot depend on a path in another repo. The harness takes the script as an argument, so it converts directly to pytest if the tool is ever adopted here.

### The "cannot be more destructive" property — checked, and it holds

A near-miss now exits **4**, naming candidates with open-row counts and the corrected command; a genuinely idle lane still exits 0. No path falls back to a broader selection, and nothing here makes `sweep` easier to reach. The change is strictly a refusal.

### The gate

```
pytest (macos-latest, py3.11): FAILURE      pytest (macos-latest, py3.12): FAILURE
ubuntu 3.11/3.12, windows 3.11/3.12, integration x2: SUCCESS
```

`tests/test_ten_lane_highway_orphan_rows.py` is new in this PR, so these are **not** pre-existing:

```
:192  assert 'drbf-compaction-notice-ab(6)' in 'drbf-compaction-notice-ab(no-such-lane)(6)'
:211  assert got["live"] == 1 and got["ended"] == 1   ->   live == 0
```

The second is the one that matters. **On macOS the liveness half of the join is not detecting a live lane** — and a join that sees zero live lanes reports a *live* lane's rows as orphaned. That is the cry-wolf direction this design was explicitly built to avoid, and it is unverified on a supported platform.

It may well be environmental (no tmux server on the runner, or a `tmux ls` output difference) rather than a logic error — but *"probably environmental"* is exactly the reasoning I declined to accept on #306, and I am not going to accept it from a lane whose work is otherwise this careful.

**Smallest paths to green, pick either:** (1) if the runner has no usable tmux, `skipif` the liveness-dependent cases with a **stated** reason, exactly as `#310` did for Windows/bash; (2) if `tmux ls` differs on macOS, parse it portably. Either way, say in the DONE-NOTE whether `highway_status.sh`'s liveness detection is intended to work on macOS at all — these scripts are GNU/Linux-only by construction (`stat -c %Y`), and if that is the answer it should be stated once, loudly, rather than rediscovered.

Everything else is merge-ready. This is one gate and the fix is small.

bkrabach:
## macOS gate cleared — and the cause was neither the tests nor the fix.

I held this PR on red macOS CI last cycle. The cause is now measured, and it is a **platform limitation the script documents about itself**:

```
highway_status.sh:53   # (stat -c %Y below is GNU/Linux; adjust for macOS if this ever travels.)
highway_status.sh:70   if [ -f "$log" ]; then age="$(( now - $(stat -c %Y "$log") ))s"; fi
```

`stat -c %Y` is a **GNU coreutils flag with no BSD/macOS equivalent**. The existing guard caught only Windows and missing-bash — but **macOS has bash**, so it never fired, and the suite ran against a script that cannot work there. That is why `live == 0` on macOS: the liveness sensor was never going to report anything.

So the honest remedy is the one applied to #310 (Windows/bash) and asked for here: **skip off GNU/Linux with the reason stated**, rather than waive silently or 'fix' a test to accommodate a script that is Linux-only by construction. Linux CI — which is where these scripts actually run — is unchanged and still green.

If the scripts are ever made portable (`stat -f %m` on BSD), delete the guard and the suite returns on macOS. The constraint is now stated once, loudly, instead of being rediscovered by the next contributor's red CI.

**All checks green.** Everything I verified last cycle stands: both fail-befores reproduce against current main (13 failed → 13 passed; the near-miss harness `RESULT: FAIL — 3 expectation(s) unmet` against the real unpatched `lane_teardown.sh`), the two halves have deliberately opposite failure policies (report fails OPEN, destroy fails CLOSED, exit 4), the real-tmux test exists and runs, subprocess envs are built explicitly per `etuz`, and no path falls back to a broader selection. Merging.

PR #315, opened 2026-09-06
fix(bundle): stop a composed behavior bundle from replacing the root's system prompt

bkrabach:
## The defect

Every session on the reporting host since 2026-08-26 ran **without the system prompt its
bundle declared** — 54 of 54 `anchors-amp-dev` sessions (08-26 ×1, 09-01 ×6, 09-02 ×18,
09-03 ×25, 09-06 ×4). Not a regression from recent work; it predates it.

`raw.system` was one block of ~98–100k chars beginning `"# Notify Bundle\n\nThis bundle
provides desktop and terminal notifications..."` — the **notify README** — followed by the
app-bundle `<context_file>` blocks. The root's own body never appeared, and neither did its
@mention.

## Root cause — three facts that compose

1. foundation `bundle/_dataclass.py` documents *"instruction: later replaces earlier"* and
   implements it: `if other.instruction: result.instruction = other.instruction`.
2. `runtime/config.py` **always** composes app-level behaviors (modes, skills,
   cli-expertise, wayfinder, notifications) and then settings' `bundle.app` list **onto**
   the user's root bundle — root is `self`, behaviors are `others`.
3. `_build_notification_behaviors()` deliberately composes notify's **root** `bundle.md`
   ("a minimal marker that just identifies the repo"), and that file carries a 2,988-char
   README body. A body *is* an `instruction`.

Net: the user's chosen bundle loses its system prompt to whichever always-composed bundle
last had a markdown body. **Notify is only the bundle that fires today** — any behavior or
app bundle with a body does this, which is why the fix is in the compose loop, not in notify.

The defect has **two halves**: the body is dropped, and *therefore* the @mention inside it is
never resolved — `mentions:resolved` listed the 22 app-bundle context mentions and not the
root's own.

## The fix, and why this one

`lib/bundle_loader/prepare.py` re-asserts the root bundle's non-empty instruction after each
behavior compose, and **warns**, naming the bundle whose body was dropped.

Option (b) — compose behaviors into a base first and the user bundle **last**, so "later
replaces earlier" favours the user — was considered and **not** taken, though it is the
better long-term shape: it matches `compose()`'s own docstring (*"typical usage:
result.compose(user_bundle)"*) and foundation's `registry.py`, which resolves `includes:`
exactly that way (`result = result.compose(included)` in a loop, then
`return result.compose(bundle)`).

The argument against it is scope, from the code: `instruction` is the **only** field in
`compose()` whose earlier value is destroyed. `session`/`spawn` deep-merge,
`providers`/`tools`/`hooks` merge by module id, `context` accumulates under a bundle prefix,
`agents` updates by name. Moving the user bundle to last inverts precedence on `session`,
`spawn`, `providers`, `tools`, `hooks`, `agents`, `name`, `version`, `description` and
`base_path` — nine surfaces, none shown to be wrong, across every app bundle a user has
configured. One broken field is repaired in that field; the reorder deserves its own change
and its own evidence.

### What now happens to a behavior bundle's body

It is **dropped, and warned about** — never silently. A silent drop in the other direction
would be the same class of defect:

```
Bundle 'notify' (git+https://github.com/microsoft/amplifier-bundle-notify@main) carries a
markdown body; dropping it. A composed behavior/app bundle must not replace the root
bundle's system instruction. Move that prose into README.md and leave bundle.md's body empty.
```

A root bundle with **no** body of its own still inherits a composed body — only a non-empty
root instruction is restored — so bodyless roots behave exactly as before
(`test_behavior_body_still_used_when_root_has_none`).

### Unit test — fails on main, passes on this branch

`tests/lib/bundle_loader/test_root_instruction_survives_behaviors.py` composes a root bundle
whose body is a single @mention with a behavior bundle whose body is README-style prose, then
drives foundation's **real** system-prompt factory (via `PreparedBundle.create_session()`
with a fake context module) to assert on the prompt and on `mentions:resolved`.

**FAIL on `main` (6afcae3)** — `docs/lanes/f26u-root-instruction-replaced/evidence/fail-before.txt`:

```
FFFF.                                                                    [100%]
E  AssertionError: the root bundle's instruction was replaced during behavior composition;
   got: '# behaviorbundle\n\nBEHAVIOR-README-BODY-f26u'
E  AssertionError: the root bundle's system.md never reached the system prompt;
   prompt was: '# behaviorbundle\n\nBEHAVIOR-README-BODY-f26u'
E  AssertionError: the dropped body was not reported; warnings were: []
FAILED ...::test_root_instruction_survives_behavior_composition
FAILED ...::test_last_behavior_with_a_body_does_not_win
FAILED ...::test_root_mention_still_expands_in_the_system_prompt
FAILED ...::test_dropped_behavior_body_is_reported_not_silent
4 failed, 1 passed in 0.09s
```

**PASS on this branch** — `evidence/pass-after.txt`:

```
test_root_instruction_survives_behavior_composition PASSED               [ 20%]
test_last_behavior_with_a_body_does_not_win PASSED                       [ 40%]
test_root_mention_still_expands_in_the_system_prompt PASSED              [ 60%]
test_dropped_behavior_body_is_reported_not_silent PASSED                 [ 80%]
test_behavior_body_still_used_when_root_has_none PASSED                  [100%]
5 passed in 0.08s
```

### Real session — the check a unit test cannot make

`amplifier run "hi" --bundle anchors-amp-dev`, same host, same `settings.yaml`, same 20 app
bundles; only `prepare.py` differs (`git stash` / `git stash pop`).

**BEFORE** (session `212e9f74`, `evidence/real-session-before.txt`):

```
chars=98406  head=# Notify Bundle\n\nThis bundle provides desktop and terminal notifications...
marker "configured for development OF" present? false
mentions:resolved -> 22 mentions, anchors-amp-dev:context/system.md NOT among them
```

**AFTER** (session `6c22e2f8`, `evidence/real-session-after.txt`):

```
chars=97895  head=@anchors-amp-dev:context/system.md
marker "configured for development OF" present? true
  -> "configured for development OF the Amplifier ecosystem itself --"
"# Notify Bundle" still present? false
mentions:resolved -> 23 mentions, INCLUDING @anchors-amp-dev:context/system.md
```

Both halves fixed: the body is back **and** its @mention resolves.

### Full suite

```
1841 passed, 1 skipped, 13 deselected, 1 xfailed in 20.97s
```

(`evidence/full-suite.txt`. The known order-dependent `test_truststore_wrap_bio_shim.py`
failure did not reproduce in this run.)

## Upstream mitigation (secondary, not a substitute)

notify's `bundle.md` body is the trigger, not the cause. Filed as draft PR microsoft/amplifier-bundle-notify#10 (Issues are disabled on that repo): move the README-style
body out of `bundle.md` into `README.md` so `bundle.md`'s body is empty. Fixing notify alone
would leave the next behavior bundle with a body to re-break this — which is why the app-cli
guard is the primary fix.

## Not in this change

- **foundation** — option (c), a warning inside `compose()` when a non-empty instruction is
  replaced, belongs to `amplifier-foundation` and is out of this lane's scope. Reported as a
  boundary rather than implemented here.
- **The compose-order alignment** described above (user bundle last, matching
  `registry.py`'s include handling) — a larger change, listed with the nine surfaces it would
  have to be measured against.

## Spend

$1.02 of API spend (two real `amplifier run "hi"` verification sessions: $0.75 before,
$0.27 after). No DTU, no infrastructure. Item authority was $0 for purchases; these are
local verification invocations, which the item names as normal.

Refs: `model_performance-f26u`


bkrabach:
## Manager verification — all gates hold. Admin-merging per owner authorization.

Head `fb8d58d`, base `6afcae3` = current `origin/main`. **CI green 9/9** across ubuntu/macOS/Windows + both integration jobs.

### Fail-before, run myself in a scratch clone

```
main   6afcae3 : 4 failed, 1 passed
branch fb8d58d : 5 passed
```

Matching the marker exactly. The failures name the defect in its own words — *"the root bundle's instruction was replaced during behavior composition"*, *"the root bundle's system.md never reached the system prompt"*, *"the dropped body was not reported; warnings were: []"*.

The **1 passed on main** is the load-bearing negative control: `test_behavior_body_still_used_when_root_has_none` passes on **both** sides, so a root bundle with no body of its own still inherits a composed body exactly as before. No phantom instruction, no regression for body-less roots.

### The fix shape — option (a), and the argument for it is better than the item's

`prepare.py::_preserve_root_instruction`, called from the `compose_behaviors` loop. The reasoning I want on the record:

> `instruction` is the **only** field in `Bundle.compose()` whose earlier value is **destroyed** rather than combined — session/spawn deep-merge, providers/tools/hooks merge by module id, context accumulates under a bundle prefix, agents update by name.

So the defect is field-specific and the repair is field-specific. Option (b) — user bundle composed last — is the better long-term shape and was **still not taken**, because it inverts precedence on nine surfaces across every configured app bundle (20 in the reproducing `settings.yaml`), none of which the evidence shows to be wrong today. That is the right call: a fix should not silently re-order things it has no evidence about.

The lane also found something the item did not carry: foundation's `registry.py` already resolves `includes:` **owner-last** (`result = result.compose(included)` in a loop, then `return result.compose(bundle)`). So (b) isn't merely the docstring's "typical usage" — it is foundation's actual convention, and app-cli's behavior loop is the deviation. Worth a follow-up in foundation, not a drive-by here.

**Behavior bodies are DROPPED and WARNED about**, naming the bundle and telling its author what to do. Not silent, and not appended — appending a 2,988-char README would re-import the exact noise, just lower in the prompt.

### Real-session verification — the check a unit test could not make

Two `amplifier run "hi"` invocations from the same worktree/venv, same host, same 20 app bundles, with `prepare.py` the **only** variable (stash / stash pop):

| | before | after |
|---|---|---|
| `raw.system` head | `# Notify Bundle\n\nThis bundle provides desktop and terminal notifications...` | `@anchors-amp-dev:context/system.md` |
| marker present | **false** | **true** |
| `mentions:resolved` | 22 (root's absent) | **23, including `@anchors-amp-dev:context/system.md`** |

Marker line, verbatim:

```
configured for development OF the Amplifier ecosystem itself --
```

**Context includes are not dropped** — the mention count goes **up** by exactly one (the root's own), so all 22 app-bundle context mentions survive. That was the failure mode I most wanted to rule out.

### Suite and lint

`1 failed, 1840 passed, 1 skipped, 13 deselected, 1 xfailed`. The one failure is `test_truststore_wrap_bio_shim.py::test_real_truststore_is_covered_at_cli_import` — **pre-existing and not this PR's**: it passes in isolation on both parent and branch (11 passed each), a collection-order dependency surfaced by adding a test file. Local ruff reports 4 errors on `prepare.py` **identically on main's unchanged copy**, i.e. version skew in my environment, not something introduced. CI's own lint gate is green.

### Notify mitigation filed upstream

microsoft/amplifier-bundle-notify#10 (draft) — `bundle.md`'s body removed, with the two things it documented that `README.md` did not (behavior-subdirectory include form, profile exclusion) moved into `README.md` so nothing is lost. **Explicitly secondary**: fixing notify alone would leave the next behavior bundle that grows a body to re-break this. Not merged — the maintainer's call.

Squash + `--admin`, owner-authorized.

PR #316, opened 2026-09-06
fix(config): compose notify's BEHAVIORS, not the notify ROOT bundle

bkrabach:
## What

`_build_notification_behaviors()` now returns **only** `#subdirectory=behaviors/*.yaml`
URIs. The bare notify **root** URI is gone.

```diff
-    # Root bundle first — a minimal marker that just identifies the repo
-    # and ensures the bundle gets cached with proper SHA metadata (fixes
-    # the "unknown" version issue during `amplifier update`). The actual
-    # functionality comes from the subdirectory behaviors below.
-    behaviors.append("git+https://github.com/microsoft/amplifier-bundle-notify@main")
-
     if flags.desktop_enabled:
         behaviors.append(
             "…amplifier-bundle-notify@main#subdirectory=behaviors/desktop-notifications.yaml"
         )
```

## Why

That "minimal marker" is not minimal: notify's root `bundle.md` carries an 82-line
README body, and `Bundle.compose()` replaces the root instruction whenever a composed
bundle has a non-empty markdown body. That is the defect #315 fixed.

**#315's guard is untouched, and stays.** `_preserve_root_instruction()`
(`lib/bundle_loader/prepare.py:71`) removed the *damage*; this PR removes the
*trigger*. Defense in depth is the point — the guard still protects every other
body-carrying bundle a user composes, including their own `bundle.app` entries, which
app-cli does not control. All five #315 guard tests pass unchanged.

## The `amplifier update` SHA motivation — preserved, and it needed no replacement

The root line existed to fix a real symptom (`2fa2c47`, "fix: resolve 'unknown' SHA
for bundles during update check"). Removing it without understanding that would trade
a fixed bug for a reintroduced one, so the symptom was reproduced first.

**The root composition was never what supplied the SHA.**
`GitSourceHandler._get_cache_path()` keys the clone on `sha256("<git_url>@<ref>")` —
the `#subdirectory=` fragment is **not** part of the key. So all three notify URIs
resolve to one cache directory:

```
ROOT     cache_path = amplifier-bundle-notify-a1c452253c0fb2e2   subpath=''
DESKTOP  cache_path = amplifier-bundle-notify-a1c452253c0fb2e2   subpath='behaviors/desktop-notifications.yaml'
PUSH     cache_path = amplifier-bundle-notify-a1c452253c0fb2e2   subpath='behaviors/push-notifications.yaml'
```

`amplifier update` resolves `notify` through `WELL_KNOWN_BUNDLES` → the ROOT URI →
`GitSourceHandler.get_status()`, which reads that same entry. Fetching a behavior
fills it.

### Fail-before / pass-after — real repo, real CLI, isolated `AMPLIFIER_HOME`

```
$ AMPLIFIER_HOME=<empty> amplifier update --check-only
│ modes   │ unknown │ 3e2a9e6 │ ● │
│ notify  │ unknown │ 7f5ff46 │ ● │      <-- symptom reproduced

# fetch ONLY the two #subdirectory=behaviors/*.yaml URIs; the ROOT is never resolved
$ AMPLIFIER_HOME=<same> amplifier update --check-only
│ modes   │ unknown │ 3e2a9e6 │ ● │      <-- CONTROL: behavior-only in code, not fetched here
│ notify  │ 7f5ff46 │ 7f5ff46 │ ✓ │      <-- real SHA, up to date
```

The `modes` control rules out "the second run just works". End-to-end through the
changed code (feed `_build_notification_behaviors()`'s own output into a fresh cache,
then run the real command) gives the same `notify │ 7f5ff46 │ 7f5ff46 │ ✓`.

Evidence: `docs/lanes/ozbw-notify-behaviors-not-root/evidence/`
(`live-update-sha.txt`, `e2e-branch-update-sha.txt`, `fail-before-pass-after-hermetic.txt`).

## Real-session check — `amplifier run "hi" -B anchors-amp-dev`

Same host, same settings (`desktop` + `ntfy` enabled), same bundle; scratch build of
each tree.

| | BEFORE (`dab0f7c`) | AFTER (this branch) |
|---|---|---|
| `raw.system` first line | `@anchors-amp-dev:context/system.md` | same |
| marker `configured for development OF` | true | true |
| `# Notify Bundle` anywhere in `raw.system` | false | false |
| `<context_file>` blocks | 23 | 23 |
| `mentions:resolved` | 23 | 23 |
| **#315 drop-warning naming `amplifier-bundle-notify`** | **1** | **0** |

`raw.system` is correct in *both* — that is #315 doing its job. The one line that
moves is the last: the root body no longer enters `compose()` at all, so the guard
has nothing to drop and users stop seeing that warning every session.
(23 = 22 app-bundle context includes + the root bundle's own `system.md`, matching
f26u's measured `mentions:resolved=23`.)

## Audit: is this a bug or a pattern?

Every always-on behavior builder in `runtime/config.py`, checked by running them:

| builder | verdict |
|---|---|
| `_build_modes_behaviors` | behavior-only |
| `_build_skills_behaviors` | behavior-only |
| `_build_wayfinder_behaviors` | behavior-only |
| `_build_app_cli_behaviors` | behavior-only |
| `_build_notification_behaviors` | **was the last root-composer — fixed here** |

**Bare-root URIs composed as behaviors: 0.** `modes` was the other one and was already
fixed in `91e1beb`; notify was the last survivor. The 6th `compose_behaviors` source is
the user's own `bundle.app` list, which is not app-cli's to police — that is exactly
what #315's guard is for. See `evidence/config-audit.txt`.

## Tests

New `tests/test_notify_behaviors_not_root.py` (+12):

1. **No bare root URI**, and every returned URI carries `#subdirectory=behaviors/` and
   ends in `.yaml` — across all three enablement combinations.
2. **Exact-list assertions** per combination.
3. **Pattern guard** — *no* always-on builder may return a fragment-less URI, so this
   cannot silently come back in a sibling.
4. **The invariant that makes the removal safe** — the behavior URIs and the root URI
   resolve to the same git cache path. If amplifier-foundation ever makes its cache key
   fragment-sensitive, this fails loudly here with an explanatory message instead of
   silently reintroducing `unknown`.
5. **End-to-end fail-before/pass-after** against a throwaway local git repo shaped like
   notify (root `bundle.md` *with* a body): nothing fetched → `cached_commit is None`
   (the "unknown"); behavior URI only → the real SHA. `skipif(win32)` with a stated
   reason (a `C:/…` drive-letter path is not expressible in `_build_git_url`'s
   host/path split); the same invariant is pinned platform-independently by test 4.

Two stale comments in `test_notification_hook_overrides_defaults.py` ("composes the
root bundle + …") corrected; no assertions changed.

### Full suite

```
BASELINE (worktree at HEAD dab0f7c):  1841 passed, 1 skipped, 13 deselected, 1 xfailed
THIS BRANCH:                          1853 passed, 1 skipped, 13 deselected, 1 xfailed
```

Delta = **+12**, all new, all passing. Reported honestly: an earlier run of *both*
trees produced one failure —
`tests/test_truststore_wrap_bio_shim.py::test_real_truststore_is_covered_at_cli_import`
(1 failed/1840 baseline, 1 failed/1852 branch) — the known collection-order dependency
that passes in isolation (11 passed). Present with and without this change; not
absorbed. See `evidence/full-suite.txt`.

## Related

`microsoft/amplifier-bundle-notify#10` empties notify's root `bundle.md` body. After
this PR the root body is inert **twice over** — never composed *and* guarded — so #10
becomes hygiene rather than a fix. Not merged here; the maintainer's call.

Generated with Amplifier

### DIFF
--- a/amplifier_app_cli/runtime/config.py
+++ b/amplifier_app_cli/runtime/config.py
@@ -1321,6 +1321,32 @@ def _build_notification_behaviors(flags: NotificationFlags) -> list[str]:
     sibling consumer ``AppSettings.get_notification_hook_overrides()`` reads
     the same flags, so the two paths cannot drift apart on defaults.
 
+    Only the behaviors, NEVER the root ``bundle.md`` — same reasoning as
+    ``_build_modes_behaviors`` / ``_build_skills_behaviors`` /
+    ``_build_wayfinder_behaviors`` / ``_build_app_cli_behaviors``, all of
+    which already compose behavior-only URIs. notify's root ``bundle.md``
+    carries an 82-line README body, and ``Bundle.compose()`` replaces the
+    instruction whenever the composed bundle has a non-empty markdown body.
+
+    The root URI used to be composed here anyway, as "a minimal marker …
+    ensures the bundle gets cached with proper SHA metadata (fixes the
+    'unknown' version issue during `amplifier update`)". It is not needed
+    for that, and it never was: ``GitSourceHandler._get_cache_path()`` keys
+    the clone on ``sha256("<git_url>@<ref>")``, which does NOT include the
+    ``#subdirectory=`` fragment. Fetching a behavior URI therefore populates
+    the exact same cache entry that ``amplifier update``'s status check for
+    the ROOT URI reads (``update.py::_check_bundle_uri`` →
+    ``GitSourceHandler.get_status``, resolving "notify" through
+    ``WELL_KNOWN_BUNDLES``). Measured on the real repo in an isolated
+    ``AMPLIFIER_HOME``: with only the two behavior URIs fetched and the root
+    never composed, ``amplifier update --check-only`` reports notify at
+    ``7f5ff46``, not ``unknown``.
+
+    That invariant is pinned by
+    ``tests/test_notify_behaviors_not_root.py`` so a future
+    fragment-sensitive cache key in amplifier-foundation fails loudly here
+    instead of silently reintroducing the "unknown" version.
+
     Args:
         flags: Resolved notification enablement.
 
@@ -1332,12 +1358,6 @@ def _build_notification_behaviors(flags: NotificationFlags) -> list[str]:
         return []
 
     behaviors: list[str] = []
-
-    # Root bundle first — a minimal marker that just identifies the repo
-    # and ensures the bundle gets cached with proper SHA metadata (fixes
-    # the "unknown" version issue during `amplifier update`). The actual
-    # functionality comes from the subdirectory behaviors below.
-    behaviors.append("git+https://github.com/microsoft/amplifier-bundle-notify@main")
 
     if flags.desktop_enabled:
         behaviors.append(
--- a/docs/lanes/ozbw-notify-behaviors-not-root/DONE-NOTE.md
+++ b/docs/lanes/ozbw-notify-behaviors-not-root/DONE-NOTE.md
@@ -1,0 +1,318 @@
+# Lane `ozbw` — compose the notify BEHAVIORS, not the notify ROOT
+
+**Item:** `model_performance-ozbw` · **Repo:** `microsoft/amplifier-app-cli` ·
+**Branch:** `lane/ozbw-notify-behaviors-not-root` ·
+**PR:** [#316](https://github.com/microsoft/amplifier-app-cli/pull/316) (READY, not draft) ·
+**Head:** `9ff57c3f01e44dd846325580299e9ce3a8f9bf15`
+
+**Outcome: branch A — RESOLVED.** All eight deliverables DONE. Nothing NOT-POSSIBLE,
+nothing blocked. Spend authority `$0` was correct and was not exceeded.
+
+---
+
+## 1. Deliverables
+
+| # | Deliverable | State |
+|---|---|---|
+| 1 | `_build_notification_behaviors()` returns ONLY behavior URIs, never a bare root; pinned by a test | **DONE** |
+| 2 | `amplifier update` SHA motivation preserved, fail-before reproduced first | **DONE** |
+| 3 | #315 guard tests still pass; guard untouched; stated in the PR body | **DONE** (5/5) |
+| 4 | Real-session check + `amplifier update --check-only` real SHA | **DONE** |
+| 5 | Audit the rest of `config.py` for the same pattern | **DONE** (0 remaining) |
+| 6 | Update notify PR #10's description; do not merge | **DONE** |
+| 7 | PR left READY with a verification comment | **DONE** ([comment](https://github.com/microsoft/amplifier-app-cli/pull/316#issuecomment-5559861684)) |
+| 8 | Full suite green, pasted in the PR body | **DONE** (1853 passed) |
+
+---
+
+## 2. The change
+
+One deletion in `amplifier_app_cli/runtime/config.py`:
+
+```diff
+-    # Root bundle first — a minimal marker that just identifies the repo
+-    # and ensures the bundle gets cached with proper SHA metadata (fixes
+-    # the "unknown" version issue during `amplifier update`). The actual
+-    # functionality comes from the subdirectory behaviors below.
+-    behaviors.append("git+https://github.com/microsoft/amplifier-bundle-notify@main")
+```
+
+plus a docstring recording *why* the removal is safe, and 12 new tests.
+
+---
+
+## 3. The hard part — the `amplifier update` SHA motivation
+
+This is deliverable 2, and the goal was right to make it the gate.
+
+### What the original bug actually was
+
+`git log -L` on the line leads to **`2fa2c47`** — *"fix: resolve 'unknown' SHA for bundles
+during update check"*. Its own message says the update check "looks for ROOT bundle cache
+entries. Since only subdirectory entries were cached, the root bundles showed 'unknown'".
+The same commit added the root URI to **both** `_build_notification_behaviors()` and
+`_build_modes_behaviors()`.
+
+**`modes` already lost its root URI**, in `91e1beb` (*"load modes behavior only, not root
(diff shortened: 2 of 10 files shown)


bkrabach:
### Verification (lane `ozbw`, re-runnable)

**Fail-before / pass-after — `amplifier update` SHA, real repo, isolated `AMPLIFIER_HOME`**

| state | `notify` row (`local │ remote │ status`) |
|---|---|
| nothing from the repo fetched | `unknown │ 7f5ff46 │ ●` ← symptom reproduced |
| **ONLY** `#subdirectory=behaviors/*.yaml` fetched, root never composed | `7f5ff46 │ 7f5ff46 │ ✓` |
| control — `modes`, behavior-only in code, not fetched here | `unknown │ 3e2a9e6 │ ●` (unchanged across both) |

Hermetic repeat against a throwaway local git repo: root status with an empty cache →
`cached_commit is None`; after fetching the behavior URI only → the repo's real HEAD,
byte-equal to the root-composed control. Both are now regression tests.

**Real-session marker line** — `amplifier run "hi" -B anchors-amp-dev`, session
`05457635-b553-4559-8815-55306e3bfd59` (AFTER) vs `10d7a2e8-1295-481f-881d-c4c5f95c92e7`
(BEFORE, baseline worktree at `dab0f7c`):

```
marker 'configured for development OF' present  : true       (both)
  the marker line                               : You are Amplifier, configured for development OF the Amplifier ecosyst…
'# Notify Bundle' present anywhere              : false      (both)
<context_file blocks                            : 23         (both)
#315 drop-warning naming amplifier-bundle-notify: 1 BEFORE  →  0 AFTER
```

**Counts**

| | baseline `dab0f7c` | this branch |
|---|---|---|
| full suite | 1841 passed, 1 skipped, 13 deselected, 1 xfailed | 1853 passed, 1 skipped, 13 deselected, 1 xfailed |
| new tests | — | +12, all passing |
| #315 guard tests | 5 passed | 5 passed (unchanged) |
| bare-root URIs composed as behaviors | 1 (`notify`) | **0** |

Earlier full-suite runs of **both** trees showed one failure,
`test_truststore_wrap_bio_shim.py::test_real_truststore_is_covered_at_cli_import`
(1 failed/1840 baseline, 1 failed/1852 branch) — the known collection-order dependency,
11 passed in isolation. Same in both trees; not introduced here, not absorbed.

Raw captures: `docs/lanes/ozbw-notify-behaviors-not-root/evidence/`.

bkrabach:
**CI: 9/9 green** — `pytest` on ubuntu/macOS/Windows × py3.11/py3.12, both `pytest -m integration` jobs, and `license/cla`.

Note for the record: the first `pytest (macos-latest, py3.11)` attempt went red at 18s. That was **not a test failure** — the job died in `Install uv` with `API rate limit exceeded` (`astral-sh/setup-uv@v5` falling back to anonymous GitHub API access); the suite never ran. Re-run passed in 30s. Checked rather than assumed, because a red macOS job on a PR that adds a `skipif(win32)` test is exactly the shape a real platform-guard bug would take. Windows py3.11 **and** py3.12 both pass, which is the check that actually exercises that guard.

bkrabach:
## Manager verification — all 7 gates re-run by me in a scratch clone. Admin-merging.

Head `04618af`, base `dab0f7c` = current `origin/main`. **CI green 9/9** (ubuntu/macOS/Windows + both integration).

**1. Behavior URIs only.** `tests/test_notify_behaviors_not_root.py` + `test_notification_hook_overrides_defaults.py`, run against both trees:

```
main   dab0f7c : 10 failed, 15 passed
branch 04618af : 25 passed
```

Failures name it directly — `test_both_enabled_composes_both_behaviors_and_nothing_else`, `test_no_always_on_behavior_builder_composes_a_bare_root_bundle`.

**2. The `amplifier update` SHA motivation — preserved, and the mechanism is real.** This was the gate that could have failed, so I checked the claim in foundation's own source rather than accepting it. `amplifier_foundation/sources/git.py:400`:

```python
cache_key = sha256(f"{git_url}@{ref}".encode()).hexdigest()[:16]
```

The key is built from **`git_url` + `ref` only** — the `#subdirectory` fragment is not part of it. So fetching `...notify@main#subdirectory=behaviors/desktop-notifications.yaml` populates **the exact cache entry** the root URI would have. No replacement mechanism was needed because none was ever load-bearing.

The lane's deviation is honest and I accept it: *"root line removed, no replacement"* is **unreachable** in the running system — the behavior URIs still fetch the repo — so it reproduced the underlying `unknown` symptom instead, live on the real repo in an isolated `AMPLIFIER_HOME` (**notify `unknown` → `7f5ff46`**), **with `modes` as an unchanged control** proving the behavior fetch is what clears it. That is a *measured* argument, not an argued-away one, which is exactly the distinction the gate required.

**3. #315's guard is intact.** `test_root_instruction_survives_behaviors.py`: **5 passed**. The guard stays; this removes the *trigger*, not the defense.

**4. Real session** — marker `configured for development OF` **PRESENT**, `# Notify Bundle` **ABSENT** from `raw.system`, context includes present, and the #315 drop-warning goes **1 → 0** (the body no longer even enters compose). `amplifier update --check-only` reports `notify 7f5ff46/7f5ff46 OK`.

**5. Pattern audit — the finding is that it was a pattern.** All **5** always-on builders are now behavior-only, **0 bare-root URIs remain** (I grepped `config.py` myself: no `amplifier-bundle-*@main` without `#subdirectory=behaviors/`), and it is pinned by a pattern test so the next one cannot be added silently.

**6.** Suite **1853 passed** vs **1841** baseline (+12 new). The known `test_truststore_wrap_bio_shim` collection-order case is pre-existing — the lane reports it failing in **both** trees and passing in isolation, and did not absorb it.

**7.** notify **#10** description updated to *hygiene, not a fix*, still draft, not merged — the maintainer's call.

Squash + `--admin`, owner pre-authorized.

PR #320, opened 2026-09-07
perf(head): apply zc6t's lean rewrite to cli-awareness.md, 398 -> 338 chars (smy5)

bkrabach:
## What this is

Applies the **already-measured** lean-head patch for this repo's one target file,
`context/cli-awareness.md`, from `amplifier-foundation` main
(**PR #372, SHA `4384805741ed7a1a8644adfd6ded9fe1ff4b4a5a`**,
`docs/lanes/zc6t-lean-head-ship/patches/context-files/04-__app_cli__-cli-awareness.md.patch`).

This is **application, not authoring** — `zc6t` diffed, fidelity-checked and
byte-measured the rewrite; this PR lands it, **re-verifies fidelity at today's head**,
and **pins it so it cannot drift back**.

**Why it is worth a PR at all:** `behaviors/cli-expertise.yaml:29` includes
`app-cli:context/cli-awareness.md`, so this text is `context.include` content —
**paid for on every request of every session**, whether or not the CLI-expert
capability is ever used. `g7h3` measured the full lean head at **−13.57% $/task,
CI [−22.27%, −4.86%]**, with all three pre-registered estimators excluding zero.

| file | stock | lean | saved |
|---|---|---|---|
| `context/cli-awareness.md` | **398 chars** (404 bytes) | **338 chars** (342 bytes) | **−60 chars** (−15.1%) |

*Characters vs bytes: `fidelity-report.json` counts characters, `wc -c` counts bytes.
Three em-dashes (3 bytes each in UTF-8) account for the difference. Same measurement,
two units — not a discrepancy.*

---

## The apply: zero fuzz, one named divergence

**`patch -p1` FAILS.** The artifact names the *installed wheel* layout, not the source layout:

```
$ patch -p1 -F0 --dry-run < 04-__app_cli__-cli-awareness.md.patch
can't find file to patch at input line 3
|--- a/amplifier_app_cli/_bundle/context/cli-awareness.md
Perhaps you used the wrong -p or --strip option?
rc=1
```

`zc6t` measured the *resolved runtime path*, which is where `pyproject.toml`'s hatch
block **puts** the file in the built wheel:

```toml
[tool.hatch.build.targets.wheel.force-include]
"context" = "amplifier_app_cli/_bundle/context"
```

`amplifier_app_cli/_bundle/` **does not exist in the source tree** — it is a build
artifact. Stripping the three build-layout components lands on the real source path:

```
$ patch -p3 -F0 --dry-run < 04-__app_cli__-cli-awareness.md.patch
checking file context/cli-awareness.md
rc=0
```

**`-F0` is load-bearing.** The hunk matched exactly; **no fuzz was used or permitted**,
per the `l4s1` precedent (*"Hunk #1 succeeded at 56 with fuzz 2"* → hand-port instead,
because fuzz is a silent placement decision). A `-p` level is a **declared path mapping**,
not a guessed placement — but it was still verified three independent ways, because a
wrong `-p` can exit 0 with confident, wrong output:

1. `patch` named the file it touched — the only `cli-awareness.md` in the repo.
2. The result is **byte-identical** to `zc6t`'s separately-shipped `.lean.md` artifact
   (`diff` clean) — a value produced by a different process than the patch.
3. Post-apply character count is **338**, matching `fidelity-report.json`'s
   `lean_chars: 338` exactly.

---

## Fidelity re-verified at TODAY'S head (not inherited)

**The file has not drifted:** today's stock is **398 characters**, exactly
`fidelity-report.json`'s `stock_chars: 398`. The repo copy and the installed wheel copy
are `diff`-identical.

**Content-word check — every word in stock, present in lean?**

```
STOCK content-words absent from LEAN:
  ['are', 'running', 'its', 'and', 'carries', 'if', 'question', 'about']
```

All eight are **function words absorbed by the rewrite** (`You are running inside` →
`You run inside`; `that carries` → `carrying`; `If the question is about *your own task*`
→ `Your own task`). **Zero rules, constraints, commands, or pointers lost.**

**Structural carriers:** fenced blocks 0→0, bullets 0→0, headings 1→1, tool-call forms
0→0, `@mentions` 0→0, backtick spans 0→0, URLs 0→0. **There was no structural carrier
here for a shrink to drop** — recorded as a checked fact, and pinned by a test.

**The four load-bearing claims all survive:** (1) an expert exists; (2) its domain is
enumerated — commands, flags, config, session machinery; (3) the docs are *authoritative*
**and** *version-matched*; (4) the negative boundary — the user's own task is **not** this
domain. **MISSING RULES: NONE**, re-derived rather than copied.

Only markdown **emphasis markers** are lost (`**Amplifier CLI application**`,
`*your own task*`). Emphasis is not a rule, and it is part of the measured saving.

---

## The pin — `tests/test_cli_awareness_lean_pin.py`

A prose file has no compiler. Unpinned, the v1 wording returns the first time someone
"improves the docs" and the saving evaporates with nothing going red. **13 tests, 5 layers:**

1. **byte-exact text** — exact anti-drift pin.
2. **character budget** — the assertion reports the delta *per request*, so growth is a
   visible decision rather than an accident.
3. **semantic invariants (×4)** — the fidelity half. A **future further shrink is welcome**;
   dropping a load-bearing claim is not. This layer survives a rewrite the byte-exact test
   would reject, so it still guards the next lean pass.
4. **v1 prose has not returned (×5)** — names the exact fragments, so a revert fails with a
   legible reason rather than an opaque diff.
5. **structural carriers + the behavior-include path** — the second catches the *other* way
   the saving is lost: the file being moved or dropped from `cli-expertise.yaml`, leaving
   the pin green while the pointer silently vanishes from the head.

**FAIL-BEFORE** (pin test run against the stock text):

```
8 failed, 5 passed in 0.06s
FAILED test_text_is_byte_exact
FAILED test_char_budget
FAILED test_semantic_invariants_survive[the negative boundary: the user's own ta]
FAILED test_v1_prose_has_not_returned[You are running inside the **Amplifier CLI application**]
FAILED test_v1_prose_has_not_returned[its commands, flags, config, and session machinery]
FAILED test_v1_prose_has_not_returned[dedicated expert that carries the authoritative]
FAILED test_v1_prose_has_not_returned[If the question is about *your own task*]
FAILED test_v1_prose_has_not_returned[that is not this domain — handle it normally]
```

**PASS-AFTER:** `13 passed in 0.01s`.

**Windows note:** the repo ships no `.gitattributes`, so a windows-latest checkout can
materialise CRLF. The test normalises newlines before comparing, so the pin is about
**content drift**, not the checkout's line-ending policy.

---

## Tests

| run | result |
|---|---|
| Full suite, this branch | **1907 passed**, 1 skipped, 13 deselected, 1 xfailed |
| Full suite, clean `main` (change stashed) | **1894 passed**, 1 skipped, 13 deselected, 1 xfailed |
| Integration job (`-m integration`) | **13 passed**, 1909 deselected |

**+13 is exactly this PR's pin test.** Nothing else moved.

**Honest disclosure — the first full-suite run had 1 failure, and it was not mine.**
`test_truststore_wrap_bio_shim.py::test_real_truststore_is_covered_at_cli_import` failed
with `status was: 'skipped: truststore is not importable'`. Rather than assume, it was
proven independent of this change: (a) stashed the change, ran that file at clean `HEAD`
→ **11 passed**; (b) restored, re-ran the full suite → **0 failed**; (c) ran a third time
→ **0 failed**; (d) `import truststore` works from the lane venv. Cause: a transient venv
state during a `uv` re-sync — the test correctly detected `truststore` momentarily absent.
**A markdown context file cannot influence truststore importability.** Recorded rather than
quietly re-run into a green.

The runs above are **local**. This PR is **DRAFT until GitHub Actions CI is green** — no
green run is implied that does not exist.

---

## Scope — verified, not assumed

`zc6t` shipped **19 patch artifacts**. The target path was read out of **every one**:

- `context-files/00,01,02,03,05,06,07,08` → `context/<name>.md` in **bundle** repos
- `context-files/04` → **`amplifier_app_cli/_bundle/context/cli-awareness.md` ← THIS REPO**
- `tool-descriptions/*.patch` → `<tool>.description` in `amplifier-module-tool-*` / bundle repos

**Exactly one of nineteen targets this repo.** Nothing else was touched; no file outside
this worktree was written.

**Out of this repo's scope, named so it is not silently assumed handled:**

- The **one REAL weakening** `zc6t` found — `edit_file`, `missing_rules: ["ALWAYS"]` — lives
  in `amplifier-module-tool-filesystem`. `zc6t` restored it in-repo at +450 chars and pinned
  it. **Not this repo's file; not claimed here.**
- **`web_search` and `mode` are verified byte-identical no-ops** (`zc6t` F2/F3), both outside
  this repo. **No file was edited here to make a count match.**
- The two hook-generated `<system-reminder>` blocks (861 + 5,350 chars) were **not**
  hand-edited. No source file in this repo produces them — that is `model_performance-z6wa`.

---

## One recommendation for the manager (not actioned — outside this lane's owned paths)

The `_bundle/` prefix in artifact 04 is a **latent trap**: it is the one artifact of the
nineteen that names a wheel path, and a lane that reaches for `-p1`, fails, then reaches
for **higher fuzz** instead of a **higher `-p`** would land the hunk somewhere plausible
and wrong. If `zc6t`'s artifact directory is ever regenerated, capturing `__app_cli__`
targets from the **source** checkout (`context/cli-awareness.md`) rather than the installed
wheel would remove it. Not fixed here — that is another repo, and this lane writes only
inside `amplifier-app-cli`.

---

**Spend: $0.00** against a **$0.00** authority (`0 runs × 0 arms × $0 / 1.00 = $0.00`).
No API calls, no DTUs, no containers, nothing registered in the infra ledger.

**DO NOT MERGE — the manager merges.** Full lane record:
`docs/lanes/smy5-patch-app-cli/DONE-NOTE.md`, evidence under
`docs/lanes/smy5-patch-app-cli/evidence/`.

Item: `model_performance-smy5`


bkrabach:
**CI green on the final head — marking ready, not merging.**

| head | run | result |
|---|---|---|
| `84e3f99` (the code change) | [34146578224](https://github.com/microsoft/amplifier-app-cli/actions/runs/34146578224) | **9/9 pass** |
| `cf3718d` (docs-only follow-up recording the above) | [34146702792](https://github.com/microsoft/amplifier-app-cli/actions/runs/34146702792) | **9/9 pass** |

Both windows legs pass on both heads, which is the live confirmation that the pin test's newline normalisation holds — a byte-exact pin on a file in a repo with no `.gitattributes` is exactly the shape that goes red on Windows for a reason unrelated to drift, and it does not.

The lane note (`docs/lanes/smy5-patch-app-cli/DONE-NOTE.md`) records the first run inline. It deliberately does **not** record its own commit's run — a note that cites the CI of the commit that adds the note regresses forever. The final head's run id is pinned here and in the lane marker instead, both of which live outside the tree and cannot move the head.

**DO NOT MERGE — the manager merges.** Item `model_performance-smy5`.

bkrabach:
## Manager verification — re-measured in a scratch clone. Merging.

Base `be25afd`, head `cf3718d`. **CI 6/6 SUCCESS** (ubuntu + macOS + Windows × py3.11/3.12).

| gate | my measurement |
|---|---|
| `context/cli-awareness.md` | **404 → 342 bytes (−62)** |
| scope | that file + `tests/test_cli_awareness_lean_pin.py` — nothing else |
| pin test | **present**, so the lean text cannot drift back |

This applies `zc6t`'s pre-measured patch `context-files/04-__app_cli__-cli-awareness.md.patch`, one of the 21 that shipped as artifacts because that lane held a single foundation checkout while its goal named 14 repos — **my goal defect, not the lane's**.

`context.include` content is paid for on **every request of every session**, invoked or not. `g7h3` measured the full lean head at **−13.57% $/task, CI [−22.27%, −4.86%]**, all three pre-registered estimators excluding zero — a program first.

Merging under the owner's Stage-1 rule: `microsoft/` repos are admin-mergeable after verification.

PR #342, opened 2026-09-15
fix: config overrides reach session.context and session.orchestrator

bkrabach:
## What changed

`resolve_bundle_config()` walked `providers`/`tools`/`hooks` at the bundle root and inside each agent, and nothing else. The context manager and orchestrator are declared at `session.context` / `session.orchestrator` as **single module entries rather than lists**, so the walk never visited them.

The consequence was total, not partial: **no** context-manager or orchestrator setting could be overridden from `settings.yaml`. Writing

```yaml
overrides:
  context-simple:
    config:
      max_tokens: 500000
```

did nothing, silently — no error, no warning, no effect.

## Why

That contradicted the rule the loop is built on, stated in `_apply_config_overrides_to_section`'s own docstring: *"overrides.\<id\>.config is keyed by module identity, not by mount location, so it must reach a module wherever it's declared."* Holding a dict instead of a list is a mount-location accident, which is exactly what that rule exists to rule out.

Adds `_apply_config_overrides_to_entry()`, the single-entry sibling, which delegates to the list version so both paths share one definition of what an override means — normalize, match by module id, deep-merge with the override winning, preserve other keys, return the original object untouched on no match.

This is what makes context-simple's new `max_tokens` / `max_tokens_fallback` settable by a user without editing a bundle.

## How to verify

9 new tests, exercising the real seam rather than a copy of its logic, including agreement between the entry and section paths.

```
2335 passed, 2 skipped, 1 xfailed
```

## Breaking changes

None — this is additive. Overrides that previously did nothing now take effect; a `settings.yaml` that already named a context/orchestrator module id under `overrides` will begin applying, which was the documented intent.

## Coordinated five-repo change

`max_tokens` in `context-simple` was documented as "Maximum context size" but implemented as a fallback consulted only when a provider published no context window. Orchestrators always pass a provider, so the knob was silently dead in production. It is now a real **cap** (default `None` = no cap); a new `max_tokens_fallback` (default 200,000) took over the fallback role.

### Merge order matters

**Merge FIRST** (independent of each other, any order):
1. `amplifier-module-context-simple` — `feat/max-tokens-cap`
2. `amplifier-module-provider-gemini` — `feat/publish-context-window`
3. `amplifier-app-cli` — `feat/session-module-config-overrides`

**Merge AFTER those three:**
4. `amplifier-foundation` — `chore/drop-dead-max-tokens`
5. `amplifier-bundle-attractor` — `chore/drop-dead-max-tokens`

Rationale: the two config repos delete `max_tokens` lines that only become safe-to-delete once context-simple's new semantics are in.

## Cross-repo verification

DTU instance `context-overflow-fix-20260915`, all 8 target behaviors PASS:

- A Gemini session's effective budget goes from **200,000** (the old fallback) to **1,011,712** (its real published window) — a **5.1x** increase.
- `max_tokens: 500000` caps to exactly 500,000.
- A cap above the model window is a no-op.
- `settings.yaml` `overrides.context-simple.config` now reaches `session.context` through the CLI's own `resolve_bundle_config`.

| Repo | Result |
|---|---|
| context-simple | 166 passed, 1 xfailed |
| provider-gemini | 330 passed; 1 pre-existing failure (`test_image_vision_integration_with_real_api`, needs a live `GOOGLE_API_KEY`, fails on `main` too) |
| app-cli | 2335 passed, 2 skipped, 1 xfailed |
| amplifier-foundation | 1938 passed, 3 skipped; 1 pre-existing failure (`test_grpc_adapter_main.py::TestVerifyModuleType::test_non_isinstance_object_with_mount_passes`, verified failing on `main` before the change) |
| amplifier-bundle-attractor | 268 passed, 2 skipped |


bkrabach:
## The five PRs in this coordinated change

**Merge FIRST — independent of each other, any order:**
1. context-simple — https://github.com/microsoft/amplifier-module-context-simple/pull/42
2. provider-gemini — https://github.com/microsoft/amplifier-module-provider-gemini/pull/48
3. app-cli — https://github.com/microsoft/amplifier-app-cli/pull/342

**Merge AFTER those three:**
4. amplifier-foundation — https://github.com/microsoft/amplifier-foundation/pull/388
5. amplifier-bundle-attractor — https://github.com/microsoft/amplifier-bundle-attractor/pull/359

The two config repos (4, 5) delete `max_tokens` lines that only become safe-to-delete once context-simple's new cap semantics (1) are in.

